PR #194, opened 2020-03-24
Fix uneccessary plugin remounting

### LINKED ISSUE #145: Navigation drawer animation affects rendering
**Description**:
At the moment we are using a "push" animation which adjusts the width of the page content, as a result. Each opening and closing of the navigation drawer sends a re-render message to the plugin asking for it re-render to adjust for the change in width. 

This causes further issues especially when tables in tabs i.e. download-cart and download-status tables reload and fetch data with each render.

An example of the issue is shown below. You can see that the data is being reloaded with opening and closing of the navigation drawer.

downloadtable

Preferrably we would want a navigation drawer which does not affect the rendering of the plugin, possibly an overlay animation.

**Acceptance criteria**:
- [ ] Overlay animation for the navigation drawer which does not affect rendering of the page contents.

louise-davies:
## Description
This fixes the plugin remounting issue described in #145, via implementing the solution I described here: https://github.com/ral-facilities/scigateway/issues/145#issuecomment-581833290. This is due to the DAaaS project preferring the current toggle-able behaviour rather than an overlay, so this just fixes the base issue of using a memoised component.

## Testing instructions
- [x] Review code
- [x] Check Travis build
- [x] Check that this actually fixes the issue by loading `datagateway-dataview` (or another plugin) and checking opening the drawer does not unmount the plugin (aka the plugin retains its state).

## Agile board tracking
Fixes #145

### DIFF
--- a/src/routing/__snapshots__/authorisedRoute.component.test.tsx.snap
+++ b/src/routing/__snapshots__/authorisedRoute.component.test.tsx.snap
@@ -3,10 +3,10 @@
 exports[`AuthorisedRoute component renders component when user logged in 1`] = `
 <div>
   <ComponentToProtect
+    dispatch={[Function]}
     loading={false}
     location="/"
     loggedIn={true}
-    requestPluginRerender={[Function]}
     store={
       Object {
         "clearActions": [Function],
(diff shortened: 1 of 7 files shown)


GoelBiju:
@louise-davies the plugin does not unmount and then mount again now and it remains the same after opening and closing the navigation drawer.

PR #227, opened 2020-05-14
ICAT anonymous autologin

### LINKED ISSUE #219: Allow ICAT authenticator to automatically authenticate users as anonymous
**Description**:
In the use case of users wanting to see open data from ICAT, currently they would first have to log in as "anonymous". This is unintuitive, and if possible we should attempt to log a user in anonymously as the page is loading, so that they can see open data without logging in. Ideally, the user would be displayed as "not being logged in", despite having a valid access token.

**Acceptance criteria**:
- [x] When using the `icat` authenticator, when a user loads the page they are automatically logged in as anonymous when possible
- [ ] When a user is logged in as anonymous, it is displayed to them as if they were unauthenticated.

louise-davies:
## Description
This PR adds code that allows the ICAT authenticator to attempt login a user automatically using the anon plugin when the app loads. This is useful for things like public data in ICAT. This functionality has attempted to be made generic so that other auth providers can potentially also implement a similar feature if necessary.

## Testing instructions
- [x] Review code
- [x] Check Travis build
- [x] Run SciGateway with the ICAT auth provider and against the scigateway-auth server and confirm it logs you in as anonymous when the app loads

## Agile board tracking
Closes #219


GoelBiju:
As mentioned and discussed with @louise-davies , there is currently a small issue where the `autoLogin` in localStorage is still set to true even after a successful login with ICAT. So at the moment, it still shows "Sign in" even after logging with e.g. ldap.

PR #230, opened 2020-05-20
Feature/auth server setting #211

### LINKED ISSUE #211: Add setting to specify auth server location
**Description**:
Currently, the auth server is pointed to via a proxy in dev mode. This makes things awkward when trying to deploy as we also have to proxy. If this was a configurable option then this would be easier and would be in line with what we're doing with `datagateway`.

**Acceptance criteria**:
- [ ] Can specify auth url link as a setting rather than relying on proxying

tswinb:
## Description
Adds a configurable auth url setting in SciGateway to remove need for proxies.

## Testing instructions

- [x] Review code
- [x] Check Travis build

## Agile board tracking
resolves #211


louise-davies:
@JoshCu could you test this on your prod install? This means you won't have to proxy

server/auth-server.js
```diff
@@ -15,6 +15,12 @@ app.use(cookieParser());
 // this would normally be an environment variable
 const jwtSecret = 'abc123456789';
 
+let authUrl;
```
louise-davies, line 18: Where is this used? I don't think any changes are necessary in this file

src/authentication/githubAuthProvider.tsx
```diff
@@ -4,8 +4,9 @@ import BaseAuthProvider from './baseAuthProvider';
 import ReactGA from 'react-ga';
 
 export default class GithubAuthProvider extends BaseAuthProvider {
-  public constructor() {
-    super();
+  public constructor(authUrl: string | undefined) {
+    super(authUrl);
+    this.authUrl = authUrl;
```
louise-davies, line 9: Surely calling `super(authUrl)` already sets `this.authUrl`, so this is superfluous?

src/authentication/icatAuthProvider.tsx
```diff
@@ -7,15 +7,19 @@ export default class ICATAuthProvider extends BaseAuthProvider {
...
 
-  public constructor(mnemonic: string | undefined) {
-    super();
+  public constructor(
+    mnemonic: string | undefined,
+    authUrl: string | undefined
+  ) {
+    super(authUrl);
     this.mnemonic = mnemonic || '';
+    this.authUrl = authUrl;
```
louise-davies, line 16: again, probably unnecessary

src/state/reducers/scigateway.reducer.tsx
```diff
@@ -217,22 +217,26 @@ export function handleAuthProviderUpdate(
...
+        payload.authUrl
+      );
       break;
 
     default:
       throw Error(
         `Unrecognised auth provider: ${payload.authProvider}, this is a development issue as there is no implementation registered for this provider.`
       );
   }
+  provider.authUrl = payload.authUrl;
```
louise-davies, line 239: again, isn't this set by the `BaseAuthProvider` constructor?

package.json
```diff
@@ -2,7 +2,6 @@
   "name": "scigateway",
   "version": "0.1.0",
   "private": true,
-  "proxy": "http://localhost:8000/",
```
louise-davies, line 5: To avoid CORS issues when booting up the server for development, I would suggest keeping the proxy config in `package.json` and setting the default auth sever URL in `settings.example.json` to `http://localhost:3000` - this makes things work with CORS in the default set up but still makes it clear that a URL is expected for the value of that setting.

Issue #354, opened 2020-07-30
Implement dark mode with user controlled toggle


tswinb:
**Description**:
Add ability for user to switch in and out of a dark mode theme. 
Part of https://github.com/ral-facilities/datagateway-userstories/issues/112 

**Acceptance criteria**:
- [ ] A toggle option in the user menu to switch into dark mode
- [ ] A dark mode with appropriate styling

PR #355, opened 2020-07-30
Add dark mode toggle #354

tswinb:
## Description
Adds ability for user to switch in and out of a dark mode theme.

Run with ral-facilities/datagateway#357 

## Testing instructions

- [ ] Review code
- [ ] Check Travis build

## Agile board tracking
connect to #354

### DIFF
--- a/src/state/actions/scigateway.actions.tsx
+++ b/src/state/actions/scigateway.actions.tsx
@@ -135,6 +135,16 @@ export const initialiseAnalytics = (): Action => ({
 
 export const configureSite = (): ThunkResult<Promise<void>> => {
   return async (dispatch, getState) => {
+    // load dark mode preference from local storage into store
+    // otherwise, fetch system preference
+    const darkModeLocalStorage = localStorage.getItem('darkMode');
+    if (darkModeLocalStorage) {
+      const preference = darkModeLocalStorage === 'true' ? true : false;
+      dispatch(loadDarkModePreference(preference));
+    } else {
+      const mq = window.matchMedia('(prefers-color-scheme: dark)');
+      dispatch(loadDarkModePreference(mq.matches));
+    }
     await axios
       .get(`/settings.json`)
       .then((res) => {
@@ -212,14 +222,6 @@ export const configureSite = (): ThunkResult<Promise<void>> => {
 
         const loadingPlugins = loadMicroFrontends.init(settings.plugins);
         loadingPromises.push(loadingPlugins);
-
-        // load dark mode preference from local storage into store
-        // otherwise, intial state is false
-        const darkModeLocalStorage = localStorage.getItem('darkMode');
-        if (darkModeLocalStorage) {
-          const preference = darkModeLocalStorage === 'true' ? true : false;
-          dispatch(loadDarkModePreference(preference));
-        }
 
         return Promise.all(loadingPromises).then(() => {
           dispatch(siteLoadingUpdate(false));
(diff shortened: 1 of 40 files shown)


src/state/actions/scigateway.actions.tsx
```diff
@@ -211,6 +213,14 @@ export const configureSite = (): ThunkResult<Promise<void>> => {
...
         loadingPromises.push(loadingPlugins);
 
+        // load dark mode preference from local storage into store
+        // otherwise, intial state is false
+        const darkModeLocalStorage = localStorage.getItem('darkMode');
+        if (darkModeLocalStorage) {
+          const preference = darkModeLocalStorage === 'true' ? true : false;
+          dispatch(loadDarkModePreference(preference));
+        }
+
```
louise-davies, line 223: This should probably either be moved to the top of the settings loading, or before even doing the `settings.json` request to prevent the light theme "flash" if a user has a preference for dark mode. Additionally, it might also be worth using [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) if there is no local storage set: see https://blog.bitsrc.io/how-to-sync-your-react-app-with-the-system-color-scheme-78c0ad00074b for info on how to use this

src/state/middleware/scigateway.middleware.tsx
```diff
@@ -49,6 +47,10 @@ export const listenToPlugins = (
   dispatch: Dispatch,
   getState: () => StateType
 ): void => {
+  const darkModeLocalStorage = localStorage.getItem('darkMode');
+  const darkModePreference = darkModeLocalStorage === 'true' ? true : false;
```
louise-davies, line 51: Same as before, might want to use `prefers-color-scheme` - perhaps this tiny bit of common code could be refactored out? Or, if the `prefers-color-scheme` change is tiny perhaps it's not worth it

src/state/middleware/scigateway.middleware.tsx
```diff
@@ -147,6 +150,17 @@ const ScigatewayMiddleware: Middleware = ((
     return store.dispatch(requestPluginRerender());
   }
 
+  if (action.type === LoadDarkModePreferenceType) {
+    next(action);
+
+    const darkModeLocalStorage = localStorage.getItem('darkMode');
+    const darkModePreference = darkModeLocalStorage === 'true' ? true : false;
+    const theme = buildTheme(darkModePreference);
```
louise-davies, line 158: Surely if the `action` is of type `LoadDarkModePreferenceType` you can just get the preference from `action.payload.preference` rather than querying localstorage?

src/state/middleware/scigateway.middleware.test.tsx
```diff
@@ -54,6 +66,13 @@ describe('scigateway middleware', () => {
     },
   };
 
+  const loadDarkModePreferenceAction = {
+    type: LoadDarkModePreferenceType,
+    payload: {
+      preference: false,
```
louise-davies, line 72: I thought the type of `LoadDarkModePreferencePayload` means this should be `darkMode: false`? Or am I misunderstanding

tswinb:
> 
> 
> The top bar goes black when in dark mode - surely this should remain coloured? It doesn't present as much as a focus point as light mode does.

@louise-davies , I was following material-ui.com on this (they change to to black).  It's quite bright otherwise, but I can change it if you think the focus point is needed?

tswinb:
To do
- [ ] check code coverage for changes after adding reading prefers-color-scheme

agbeltran:
@tswinb now checking accessibility standards

louise-davies (approved):
Looks good!

PR #431, opened 2020-09-21
enable datagateway logos in sidebar #301

### LINKED ISSUE #301: Add logo
**Description**:
Add logo to UI

**Acceptance criteria**:
- [ ] add SciGateway logo

patrick-austin:
## Description
Logos are passed from plugins to SG with the `RegisterRoutePayload`, specifically we have three strings: `logoLightMode`, `logoDarkMode`, and `logoAltText`. The former are included in the side menu when depending on whether we are in `darkMode` or not, and `logoAltText` is added to the `displayName` when the image src is missing so that it has consistent styling.  

Rather than complex logic needed to find the host in SciGateway, we now expect this to be in the plugin's own settings as we would need it there anyway for displaying images _within_ the plugin (for example the STFC logo used in the branding header ral-facilities/datagateway#547). As a result this PR is dependent on ral-facilities/datagateway#562 which registers plugins where we have access to the settings.

## Testing instructions
To test this, ral-facilities/datagateway#562 will need to be checked out for the plugins, and they will need to be served through SG.
- [x] Review code
- [x] Check Travis build
- [x] Check the dark mode appropriate logo displays in the sidebar when pluginHost is specified in the settings (and matches the port the plugin is served on, usually 5001)
- [x] Check that no logo displays in the sidebar when pluginHost is NOT specified in the settings and if logoAltText is specified then this is included in the ListItemText

## Agile board tracking
closes #301, connect to ral-facilities/datagateway#262


agbeltran:
Discussed at sprint meeting to consider re-working this so that plugins' logos are not included in parent web app. @patrick-austin @louise-davies

src/state/reducers/scigateway.reducer.tsx
```diff
@@ -92,12 +94,62 @@ export function handleConfigureStrings(
...
+        section: '',
+        link: '',
+        displayName: '',
+        order: -1,
+      };
+      existingPlugins.push(newPlugin);
+    }
+  });
+  return existingPlugins;
+};
```
louise-davies, line 124: Why were these changes needed? Was there a problem of duplicate plugins or something? I can't fathom why these changes are needed for the logo change unless it's fixing a separate problem.

src/navigationDrawer/navigationDrawer.component.tsx
```diff
@@ -75,8 +81,16 @@ class NavigationDrawer extends Component<CombinedNavigationProps> {
...
       >
+        <img
+          className={this.props.classes.menuLogo}
+          alt={plugin.logoAltText}
+          src={
+            plugin.src &&
+            plugin.src +
+              (this.props.darkMode ? plugin.logoDarkMode : plugin.logoLightMode)
+          }
+        />
```
louise-davies, line 92: We should probably make this `img` tag conditional on the plugin actually providing the logo options, otherwise we'll get an empty image tag (which isn't particularly a problem, just a little strange).

src/state/reducers/scigateway.reducer.tsx
patrick-austin: The problem I was having was the `src` string for the image that the DG plugin registers (`plugin.logoDarkMode`/`plugin.logoLightMode`) doesn't have the full address, it's something like `/static/media/datagateway-logo.808b0870.svg` but to display the image in Scigateway I needed the `http://localhost:5001` bit as well. The only place I could find to get this was when we read SG's `settings.json` file, and then add the address to the plugin so when we display the image we can combine the two.

I wasn't sure if we registered the plugins or read the settings first, so I put in logic to handle it either way round. If the plugin isn't registered it created a placeholder with just the name and `http://localhost:5001` or whatever, then we actually register the plugin later it adds the rest of the information to that entry in the list of plugins. Or if the plugin is already registered, we just set the `src` field on the existing entry.

Like I said this was the only way I could find of getting the full address needed for hosting the image. If there's a better way of getting this to work I'd be more than happy to change it; I agree that it's not the nicest thing to look at.

src/navigationDrawer/navigationDrawer.component.tsx
patrick-austin: The reason I left it in was so the alt text (i.e. DataGateway) would display if the image failed to load (though obviously if neither `src` nor `alt` is set then it's worthless). But to be honest the formatting on that wasn't great anyway, it'd probably be better to include that in the `<ListItemText>` so it's consistent with the rest of the display name in the event we don't have the image? Either way I'll make the  conditional like you say

src/navigationDrawer/navigationDrawer.component.tsx
```diff
@@ -75,9 +96,12 @@ class NavigationDrawer extends Component<CombinedNavigationProps> {
         id={`plugin-link-${plugin.link.replace(/\//g, '-')}`}
         button
       >
+        {imgSrc && (
+          <img className={this.props.classes.menuLogo} alt="" src={imgSrc} />
```
louise-davies, line 100: Shouldn't `alt` have `logoAltText` set here?

patrick-austin: I didn't set `alt={logoAltText}` as the formatting of the image alt text (when it displays) and the rest of the text for the list item didn't match, instead I just added `logoAltText` to the start of `plugin.displayName` when a `src` for the image couldn't be found. Though thinking about it now I suppose this wouldn't cover the case where we do provide an image src, it just fails to load, or any accessiblity effects of having `alt` set. Do you think it would be better not to complicate things and just set `alt` with `logoAltText` and leave `plugin.displayName` as is?

louise-davies: I was more thinking about the accessibility implications even if the logo is displayed. I kind of like the use of logoAltText to be prepended to the display name if there's no logo though.

src/navigationDrawer/navigationDrawer.component.tsx
```diff
@@ -67,6 +73,21 @@ ForwardRefLink.displayName = 'ForwardRefLink';
...
+        ? plugin.logoLightMode
+        : undefined;
+
+    const displayText = plugin.displayName
+      ? imgSrc
+        ? plugin.displayName
+        : plugin.logoAltText
+        ? plugin.logoAltText + plugin.displayName
+        : plugin.displayName
+      : plugin.plugin;
```
louise-davies, line 89: This is quite confusing and hard to mentally process - I had to use this: https://converter.website-dev.eu/ to help me understand it! Maybe we should ignore prettier a little bit here and add some "unnecessary" brackets or indenting, or this could be refactored.

louise-davies: Maybe this works a little better?

```
plugin.displayName
      ? imgSrc || !plugin.logoAltText
        ? plugin.displayName
        : plugin.logoAltText + plugin.displayName
      : plugin.plugin;
```

patrick-austin: I simplified the logic for setting `imgSrc` - we don't need to explicitly check whether `plugin.logoDark/LightMode` is thruthy or undefined since we'd be setting `imgSrc` to be undefined anyway (and in the event it's an empty string, the later checks on `imgSrc` will behave the same as if it were undefined).

Split the `displayText` assignment into a variable for the prefix (either `logoAltText` or `''`) and then `displayName` or `plugin` for the rest of it.

Hopefully by removing the nested ternary operators makes it easier to follow (it looks better to me at least).

louise-davies (approved):

Issue #427, opened 2020-09-17
Design front-page


agbeltran:
**Description**:
Currently a static page, we need to consider design and content.

**Acceptance criteria**:
- [ ] Information about what functionality is available

louise-davies:
@agbeltran mentioned wanting data to be visible when the user visits the service, like in TopCAT. The way I would imagine this could be done is to add a configuration option "start_url" (or similar), which SciGateway redirects to once the plugins have finished loading. This fits within the current plugin based architecture quite well.

Perhaps, it could be done as part of "listening" to the plugins registering routes? (aka listen and see if any match "start_url")

agbeltran:
@agbeltran to provide text

sam-glendenning:
@agbeltran criteria includes "Information about what functionality is available". I presume this is going to vary based on what datagateway plugins are loaded. And later on what plugins and features are available for other projects outside of datagateway. This is all available in some form in the navigation bar on the left. Should we be looking to move some of that onto the homepage itself?

I also need a greater understanding of what type of content individual users may be seeing depending on if they're from diamond, isis, etc. Does this mean an isis user should be seeing an entirely different homepage than a diamond user? Perhaps not seeing certain plugins listed at all. Basically not sure what the plan is for this or if something like this is currently implemented (plugins for one type of user, not another).

agbeltran:
@sam-glendenning yes, you are right that the intention is that each plugin may have its own static homepage. We now want to focus on getting the datagatway homepage done, describing DG functionality. We had discussed with @louise-davies that we can configure the homepage per plugin, the same way that now scigateway has a configuration for the first page that is loaded.

At the moment, we have this:
![homepage-design](https://user-images.githubusercontent.com/1417033/114033229-fb057800-9874-11eb-9232-db5a705a26e1.png)

I notice that there is a need to scroll the page to see all the content, and ideally we want to avoid the need to scroll.

agbeltran:
For DataGateway, I would like to see the DG logo and I suggest the following text
- Explore: Browse, explore and visualise the experimental data produced at the facility
- Discovery: Search for the experimental data according to different criteria
- Download: Retrieve the experimental data using a variety of download methods.

We will discuss specific changes, and addition of logos, with each of the facilities, so we need to design it so that we may have different static home pages per facilty.

agbeltran:
For now, remove the text "transofrming... our national facilities.

agbeltran:
[...]

PR #447, opened 2020-10-06
allow optional redirect to plugin from homepage #427

patrick-austin:
## Description
Added `startUrl` to `settings.json` as an optional setting. When provided, if one of the plugins has the same `link` as `startUrl`, and the current path is the homepage (`/`), then a redirect to the plugin will be triggered.

If the user needs to login, the referring path will be the plugin link, so after the login redirect the plugin will be loaded.

If the user is already logged in, then a redirect is needed to ensure the plugin renders the updated path (and not what it would show for a path of `/`. This happens at the same place as the login redirect (if the `startUrl` is enabled).

If `startUrl` is not set, or no plugin matches the url, then nothing should happen (previous behaviour maintained).

## Testing instructions

- [x] Review code
- [x] Check Travis build
- [x] Check that navigating to the homepage without a startUrl specified has no change
- [x] Check that when navigating to the homepage with the startUrl and corresponding plugin enabled, you are redirected to the plugin
- [x] Check that clicking the home (SciGateway) button does not trigger a redirect and still allows access to the normal homepage
- [x] Check that navigating to a specific page (e.g. `/cookies` doesn't trigger a redirect)

## Agile board tracking
connect to #427, ral-facilities/datagateway#431

### DIFF
--- a/src/routing/authorisedRoute.component.tsx
+++ b/src/routing/authorisedRoute.component.tsx
@@ -10,6 +10,7 @@ interface WithAuthStateProps {
   loading: boolean;
   loggedIn: boolean;
   location: string;
+  startUrlState?: StateType;
 }
 
 interface WithAuthDispatchProps {
@@ -27,6 +28,7 @@ const mapStateToProps = (state: StateType): WithAuthStateProps => ({
     isStartingUpOrLoading(state.scigateway.authorisation),
   loggedIn: state.scigateway.authorisation.provider.isLoggedIn(),
   location: state.router.location.pathname,
+  startUrlState: state.router.location.state,
 });
 
 const mapDispatchToProps = (dispatch: Dispatch): WithAuthDispatchProps => ({
@@ -43,6 +45,7 @@ export default function withAuth<T>(
         loading,
         loggedIn,
         location,
+        startUrlState,
         requestPluginRerender,
         ...componentProps
       } = this.props;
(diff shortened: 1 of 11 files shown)


src/routing/authorisedRoute.component.tsx
```diff
@@ -57,8 +60,16 @@ export default function withAuth<T>(
...
           ) : null}
+          {/* If using a plugin as the start page, redirect here so the plugin renders with the redirected url */}
           {!loading && loggedIn ? (
-            <ComponentToProtect {...(componentProps as T)} />
+            startUrlState && startUrlState.scigateway.startUrl ? (
+              <Redirect
+                push
+                to={{ pathname: startUrlState.scigateway.startUrl }}
+              />
+            ) : (
```
louise-davies, line 70: Why is this bit needed? If I comment this out it still redirects to the plugin with the correct URL?

patrick-austin: When I tested this, navigating to `/` while logged in redirects to the right url (e.g. `/browse/investigation`) and it loads the plugin, but the page that's rendered isn't the page that matches the current url, but the previous url (`/`). So for this example, it displays the "Browse investigations" link, as that's what the plugin routing renders: https://github.com/ral-facilities/datagateway/blob/master/packages/datagateway-dataview/src/page/pageTable.component.tsx#L41
Adding in this redirect caused it to re-render and show the table as expected. As I recall, search and dataview display links in this way, but I don't know what download does off the top of my head.
If this isn't an issue on your end for whatever reason, and the page renders as expected rather than showing that link, then there wouldn't be any reason to include this redirect.

louise-davies: Ah, I was testing this with just the demo plugin rather than one of the datagateway plugins - after testing with dataview you're indeed correct. I assumed it was needed for a reason - thanks for the explanation :)

louise-davies (approved):

src/routing/authorisedRoute.component.tsx
patrick-austin: No problem, I could remember there was something that happened differently depending on whether the user was logged in or not when I first saw the comment, but it took me a bit of poking around to remind myself exactly what the problem was.

Issue #479, opened 2020-11-19
External links at bottom of TopCAT are needed


louise-davies:
**Description**:
Currently, TopCAT links to the facility home page, privacy policy, cookie policy and about us pages at the bottom of the screen. Currently, SciGateway only has the cookie policy page. These other pages (which are external links) must be added and be configurable.

Suggestions on the best place to put these are welcome. Normally it's at the bottom of the page, so perhaps that's where these should go too? Cookie management can stay where it is though I think.

**Acceptance criteria**:
- [ ] Required external links are displayed on the page somewhere
- [ ] The urls can be configured

VKTB:
Think I've misunderstood this issue, originally I thought that these links will be the same and need to be displayed on every page on the website, but then I came across Diamond's TopCat and realised that the links are different.

@louise-davies Should these links be only displayed on the plugin pages and no other pages on the website?

I also think that they should be displayed at the bottom of the page.

louise-davies:
They probably need to be displayed on all pages of the site, as they're there for policy reasons. The bottom of the page makes sense, as that's where most other websites have them. The links have to be configurable because the links for ISIS will differ from those for Diamond.

What do you mean you came across Diamond's TopCAT and realised the links are different? Do you mean that the urls go to different places/have slightly different text? Because that's what we need to replicate. I took a look how TopCAT does it (see https://github.com/icatproject/topcat/blob/master/yo/app/languages/en_dls.json#L476) and it seems like it just allows raw HTML in the string file (which I suppose is ultimate configuration)

VKTB:
>What do you mean you came across Diamond's TopCAT and realised the links are different?

I meant that the links differ - the _Privacy Policy_, _Cookie Policy_ and _About Us_ links in Diamond's TopCAT are different to the one of ISIS's. And yes the name of the link that leads to the home page of the facility also differs. Initially I thought they are the same.

My understanding of SciGateway so far is that it is a single interface that allows for data from different facilities to be accessed by users. As a result, the _Home_, _Contact_, _Sign In_ (and possibly the _Download_ and _Search_) pages are not specific to any of the facilities (ISIS or DLS)? If so, I think that the above pages should not display these links.

louise-davies:
[...]

PR #492, opened 2020-11-25
Add footer to display external links at bottom of pages

VKTB:
## Description
Adds a footer component that displays external links that direct the user to the facility home, cookie policy and about us pages. To allow facilities to configure these external links and their names, a raw HTML is stored in in the `default.json` file. I noticed that the link names change colour after they have been clicked (which is normal behaviour), so I just wanted to get people's view on whether we want to override this? It seems that is overridden in TopCAT and `text-decoration` is set to `none` to remove the lines under the link names. Given that there are some hyperlinks in the _Cookies_ page, it might better to globalise this if we want the same behaviour there.

The footer component is not tested as I wanted to check whether what I have done so far meets the requirements before writing any tests.

image

## Testing instructions
- [ ] Review code
- [ ] Check Travis build
- [ ] Change the `html` value of `footer` inside `default.json` to 
```
<a href=\"http://www.diamond.ac.uk\">Diamond Light Source</a> | <a href=\"http://www.diamond.ac.uk/Legal.html\">Privacy Policy</a> | <a href=\"http://www.diamond.ac.uk/Home/About.html\">About Us</a>
``` 
or a raw HTML value of your own choice.
- [ ] Check that the external links are displayed at the bottom of the pages.
- [ ] Check that you are successfully redirected to the external web page when you click on an external link.
## Agile board tracking
connect to #479

### DIFF
--- a/public/res/default.json
+++ b/public/res/default.json
@@ -46,5 +46,8 @@
     "link-text": "Learn more",
     "accept-button": "Accept",
     "manage-preferences-button": "Manage Preferences"
+  },
+  "footer": {
+    "html": "<a href=\"#\">Facility Name</a> | <a href=\"#\">Privacy Policy</a> | <a href=\"#\">About Us</a>"
   }
 }
--- a/src/__snapshots__/pageContainer.test.tsx.snap
+++ b/src/__snapshots__/pageContainer.test.tsx.snap
@@ -10,5 +10,6 @@ exports[`PageContainer - Tests renders correctly 1`] = `
   <Connect(WithTheme(Tour)) />
   <Connect(WithStyles(CookieConsent)) />
   <Connect(WithStyles(Routing)) />
+  <Connect(WithStyles(Footer)) />
 </div>
 `;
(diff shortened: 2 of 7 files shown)


louise-davies (changes requested):
Works well with the plugins, but I think there is too much whitespace around the footer and it takes up too much screen space because of it. Also the font size could also be smaller - again to free up some screen space. Additionally, we'd probably want to reduce the size of the plugins to not need to scroll to the footer, but that is a separate issue.

VKTB:
@louise-davies Reduced the whitespace and the font size as requested, see screenshots below.

image
---
image

louise-davies (approved):
Looks good - there'll probably need to be a separate issue opened in DataGateway to adjust the plugin container sizes to account for the footer without scrolling - I'll leave that to you @VKTB ?

VKTB:
@louise-davies I opened an issue in DataGateway and will provide a fix for it soon. What do we want to do about the below?

>I noticed that the link names change colour after they have been clicked (which is normal behaviour), so I just wanted to get people's view on whether we want to override this? It seems that is overridden in TopCAT and `text-decoration` is set to `none` to remove the lines under the link names. Given that there are some hyperlinks in the _Cookies_ page, it might better to globalise this if we want the same behaviour there.

VKTB:
The links were difficult to see in dark mode and therefore failed accessibility. I overrode the colours of `<a>` element and they are consistent with the colours used for the links in the cookies page.

Before:
image

After:
image
image

louise-davies (approved):

PR #529, opened 2021-01-08
Add Contact page

### LINKED ISSUE #527: Add Contact page
**Description**:
A Contact page would be needed to provide contact details to users.

**Acceptance criteria**:
- [ ] Static Contact page is created and can be accessed
- [ ] Page text is configurable in some way

VKTB:
## Description
Adds a static contact page whose text can be configured from within `default.json`.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Static Contact page is created
- [ ] Contact page can be accessed when the _Contact_ button from the main AppBar.
- [ ] Page text is configurable

## Agile board tracking
Closes #527


src/contactPage/contactPage.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { StateType } from '../state/state.types';
+
+const styles = (theme: Theme): StyleRules =>
+  createStyles({
+    root: {
+      padding: theme.spacing(2),
+      backgroundColor: theme.palette.background.default,
+      '& a': {
+        '&:link': {
+          color: '#1E5DF8',
```
patrick-austin, line 22: Could replace this with `(theme as UKRITheme).ukri.bright.blue` (and `purple`/`red` below).

patrick-austin (changes requested):
Works fine as it is, but could avoid hardcoding the hex for the hyperlink colours by importing the `UKRITheme` (see comment).

Should also replace the [hardcoded styling](https://github.com/ral-facilities/scigateway/blob/feature/add_contact_page-%23527/public/res/default.json#L38) in `default.json` for the cookies info with what we're doing here (i.e. using `createStyles` and `'& a'`). I don't think I knew how to apply styles to child elements back when I wrote that one, so that's my bad.

src/contactPage/contactPage.component.tsx
VKTB: Wasn't aware that the theme existed but I like this idea.

patrick-austin (approved):
Looks great, and thanks for sorting out the cookies page as well.

PR #532, opened 2021-01-12
Add Help page

### LINKED ISSUE #478: Add more traditional help page
**Description**:
In DLS demo, it was suggested that a more traditional help page would be needed, especially to explain the differences to existing users of TopCAT.

**Acceptance criteria**:
- [ ] Static help page is created and can be accessed
- [ ] Page text is configurable in some way

VKTB:
## Description
Adds a static Help page whose text and any images can be configured from within `default.json`. To allow for reviewing, I purposely used the text and images from TopCat's Help page. I will open an issue so that the page content can be updated once the new one is ready. 

Commit [26f5ddc](https://github.com/ral-facilities/scigateway/pull/532/commits/26f5ddc7cc544d724bef032770070721a3bc402c) increases the code coverage of the Contact page component - this should have been part of #529 .

## Testing instructions
- [ ] Review code
- [x] Check Actions build
- [ ] Static Help page is created
- [ ] Help page can be accessed when the _Help_ button from the main AppBar.
- [ ] Page text is configurable

## Agile board tracking
Closes #478


patrick-austin (changes requested):
Placeholder page looks good, the only (very minor) thing I think might help is adding 
`"showHelpPageButton": true` to [the example settings file](https://github.com/ral-facilities/scigateway/blob/feature/add-help-page-%23478/public/settings.example.json#L3) alongside `showContactButton`. When `features` are provided but `showHelpPageButton` isn't a key then the button won't be present, so if I were to just copy the example settings file as is I would get the contact button but not the help page. It's only when `features` is entirely absent from the settings file that the [default value of true](https://github.com/ral-facilities/scigateway/blob/feature/add-help-page-%23478/src/state/reducers/scigateway.reducer.tsx#L61) would be used.

VKTB:
@patrick-austin Thanks for pointing this out, I will update the examples file as requested.

patrick-austin (approved):

Issue #515, opened 2020-12-14
Add support for admin user to block user access easily and present a holding page during interventions


agbeltran:
**Description**:
During interventions, an admin user must be able to block user access and presented a holding page. This part of a maintenance mode for SciGateway and the plugings. We need to discuss how this is implemented here and in combination with the plugins.

**Acceptance criteria**:
- [ ] There is a way for an admin user to block user access and present a holding page during interventions

PR #565, opened 2021-02-16
Feature/display maintnenace page if site under maintenance #515

VKTB:
## Description
Fetches the maintenance state from `scigateway-auth` and loads it into the store. If the `show` boolean flag is `true`, it displays the Maintenance page when a user tries to access any of the plugins. The Login and the other SciGateway pages can still be accessed to allow admin users to login and access the Admin page (from where they will be able to set scheduled maintenance/ maintenance) that will be created soon. Also checks for changes in the state every hour and updates the store if it detects any. This ensures that the maintenance page gets displayed when a user does not reload the site for longer than an hour but maintenance was enabled in the meantime.

I had to move the code that deals with loading the dark mode preferences in `src/state/actions/scigateway.actions.tsx` from the top to near the bottom of the function because it causes a re-render which results in the `render` function inside `routing.component.tsx` getting executed before the maintenance state is fetched. I tried putting the logic that deals with fetching the maintenance state at the top of the function but it does not work because it needs the auth provider and this does not get loaded until the values from the `settings.json` file get fecthed. Hopefully, moving this to near the bottom of the function does not cause any undesired effects.

For some reason, actions are still sent when navigating to the DataView plugin while the site is under maintenance. This does not happen with the Download and Search plugin. I spent some time debugging but I could not figure out what's causing this.

## Testing instructions
You will need to pull the latest changes from `master` in the `scigateway-auth` repo and deploy it locally. You can use Postman to set the maintenance state values but it's easier to just directly edit the `maintenance.json` file in the `maintenance` directory inside the `scigateway-auth` repo.

- [x] Review code
- [x] Check Actions build
- [x] No maintenance page is displayed when the `show` boolean flag returned by `scigateway-auth` is `false`.
- [x] No maintenance page is displayed when the `show` boolean flag returned by `scigateway-auth` is `true` and the user navigates to the Login or any of the SciGateway pages.
- [x] Maintenance page is displayed when the `show` boolean flag returned by `scigateway-auth` is `true` and the user tries to navigate to any of the DataGateway plugin pages. Check that the message displayed in the maintenance page is the same as what's returned by the backend.

## Agile board tracking
connect to #515


VKTB:
I am not sure how to unit test the logic that deals with checking for changes in the maintenance state every hour. If anyone has any ideas then please let me know.

src/routing/routing.component.test.tsx
```diff
@@ -94,6 +95,36 @@ describe('Routing component', () => {
...
+        displayName: 'Test Plugin',
+        order: 1,
+      },
+    ];
+    const wrapper = mount(
+      <Provider store={mockStore(state)}>
+        <MemoryRouter
+          initialEntries={[{ key: 'testKey', pathname: '/test_link' }]}
+        >
+          <MaintenancePage classes={maintenancePageClasses} />
```
patrick-austin, line 120: Should this be `<Routing ... />` like in the above test in order to test the logic in `routing.component.tsx`? We already test the `<MaintenancePage />` component renders correctly in `maintenancePage.component` right?

src/routing/routing.component.test.tsx
```diff
@@ -7,6 +7,7 @@ import { initialState } from '../state/reducers/scigateway.reducer';
 import { createLocation } from 'history';
 import { MemoryRouter } from 'react-router';
 import { Provider } from 'react-redux';
+import MaintenancePage from '../maintenancePage/maintenancePage.component';
 
 // this removes a lot of unnecessary styling information in the snapshots
 jest.mock('@material-ui/core/styles', () => ({
```
patrick-austin, line 13: Mocking the styles here causes trouble when rendering the MaintenancePage component, but we can mock that too:
`jest.mock('../maintenancePage/maintenancePage.component');`
The snapshot then contains `Connect(MaintenancePage)` rather than `Connect(WithAuthComponent)` (which it contains when maintenance mode is false). This covers all of `routing` but looks like it reduces coverage in `maintenancePage` despite the fact the local unit tests don't mark any lines as uncovered. So not sure what's going on there.

src/maintenancePage/maintenancePage.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+    </div>
+  );
+};
+
+const mapStateToProps = (state: StateType): MaintenancePageProps => ({
+  message: state.scigateway.maintenance.message,
+});
+
+export const MaintenancePageWithoutStyles = MaintenancePage;
+export const MaintenancePageWithStyles = withStyles(styles)(MaintenancePage);
```
patrick-austin, line 60: We never import `MaintenancePageWithoutStyles`, and `MaintenancePageWithStyles` is only used in tests. So we can remove the former, and shouldn't export the latter.

src/maintenancePage/maintenancePage.component.test.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { createMount } from '@material-ui/core/test-utils';
+import {
+  CombinedMaintenancePageProps,
+  MaintenancePageWithStyles,
```
patrick-austin, line 5: By importing this version rather than the connected default export we don't test mapping the store to the components props. It would be better (in terms of coverage) to import `default as MaintenancePage` and then create a `testStore` for the test rather than directly passing in `message` as a prop.

patrick-austin (changes requested):
Some changes for the tests in the individual comments, and I haven't tested it myself but https://jestjs.io/docs/en/timer-mocks says that their `jest.useFakeTimers()` works with `setInterval()`, and we use it elsewhere in the codebase already.

As for the dataview requests being sent we discussed this virtually, and it seems like it would be easier to prevent that in datagateway rather than here so that's fine for this PR.

src/routing/routing.component.test.tsx
VKTB: It was erroring when I used `<Routing ... />` so I thought that `<MaintenancePage />`. However, mocking the `MaintenancePage` component as per your suggestion in the other comment seems to work.

src/maintenancePage/maintenancePage.component.test.tsx
VKTB: That makes sense, thanks!

VKTB:
>As for the dataview requests being sent we discussed this virtually, and it seems like it would be easier to prevent that in datagateway rather than here so that's fine for this PR.

I opened the [Do not send requests if dataview element is not present](https://github.com/ral-facilities/datagateway/issues/574) issue in the DataGateway repo for this.

>Some changes for the tests in the individual comments, and I haven't tested it myself but https://jestjs.io/docs/en/timer-mocks says that their jest.useFakeTimers() works with setInterval(), and we use it elsewhere in the codebase already.

I looked at how we do it in other places in the codebase and also tried different suggestions from the net but I could not get the code inside `setInterval` to execute. I couldn't find an example where people do similar thing inside the `App` component so it might be to do with how the component works? I modified the code inside `setInterval` to only print to the console and that didn't execute either. I also moved `jest.useFakeTimers()` around a few times but no luck..

patrick-austin (commented):
The changes look good, I had a more in depth look at the interval thing and got it into the tests, but I did have to move the code into the App class (think this is why mocking the timers didn't do anything, as the code to set the interval was running before we mocked the `setInterval` function.

Moving it into `componentDidMount` still works as I'd expect, but obviously if there's anything wrong with doing it that way you can see then we can leave it out (it's just for the test after all). I'll try and push my commit to this branch so you can have a look at it, and then if you're happy with that I can approve the whole PR? At any rate let me know what you think.

VKTB:
@patrick-austin Oh, I now see what was going on there, thanks for explaining and sorting it out. I ran a few manual tests and it seems to work as expected, so I am happy with the changes that you made there.

patrick-austin (approved):

PR #591, opened 2021-02-25
Add Admin page

### LINKED ISSUE #514: Add support for admin to add message for notifying users about scheduled intervention
**Description**:
When an intervention is scheduled, an admin user needs to set up a message to notify the users about the intervention. We should discuss the responsibilities of SciGateway and the plugins to implement this functionality. 

**Acceptance criteria**:
- [ ] An admin user can set up a message to notify users about an intervention

### LINKED ISSUE #515: Add support for admin user to block user access easily and present a holding page during interventions

### LINKED ISSUE #548: Add Admin page
**Description**:
An Admin page would be needed for admin users to access and set scheduled maintenance or maintenance. Other features could be included in this page if required.

**Acceptance criteria**:
- [ ] An Admin page is created and can be accessed by admin users only

### LINKED ISSUE #602: Windows / Yarn / Git Bash / Husky compatability
**Description**:
Updating to Husky 5 works for powershell, but commiting with git bash fails. See https://typicode.github.io/husky/#/?id=yarn-on-windows for the error and workaround. 

**Acceptance criteria**:
- [ ] Git bash can be used on windows to make commits

VKTB:
## Description
Adds an admin page that allows admin users to set scheduled maintenance and maintenance. In addition to checking whether the user is logged in and is admin, the JWT token is also verified before access is granted to the Admin page. I decided to do the verification during mounting of the `authorisedRoute` component so that access, not only to the admin page component but other components that require users to be logged in, be granted only if the token is valid. If there are any reasons as to why this should not be done with other components, then please let me know and I will move the verification to the admin page component. Verifying the token before granting access to the admin page for example, ensures that unauthorised users that have tampered their token so that `userIsAdmin` is `True` cannot access the admin page.

Also fixed the husky 5 pre-commit issue that I started facing when trying to create commits using git bash.

## Testing instructions
You will need to pull the latest changes from `master` in the `scigateway-auth` repo and deploy it locally. A user can be given admin privileges by adding their username to the `admin_users` list in the `config.json` file in `scigateway-auth`. For example, if you are logging in as root then adding `simple/root` to the list will do the job.

- [x] Review code
- [x] Check Actions build
- [x] Admin page is created
- [x] _Admin_ button is added to the main AppBar if user is admin
- [x] Admin page can be accessed when the _Admin_ button from the main AppBar is clicked (user needs to be admin)
- [x] Scheduled maintenance can be set from the Admin page
- [x] Maintenance can be set from the Admin page
- [x] User gets logged out and _Page not found_ is displayed when navigating to `http://localhost:3000/admin` after the token in the local storage has been tampered so that `userIsAdmin` is `True`. You need to login in before tampering the token.

## Agile board tracking
closes #548 
closes #514 
closes #515 
closes #602


src/adminPage/adminPage.component.tsx
```diff
@@ -0,0 +1,232 @@
...
+      color: theme.palette.secondary.main,
+      fontWeight: 'bold',
+    },
+    paper: {
+      display: 'flex',
+      flexDirection: 'column',
+      marginTop: theme.spacing(2),
+      marginBottom: theme.spacing(2),
+      padding: theme.spacing(2),
+      [theme.breakpoints.up(800 + theme.spacing(4))]: {
```
patrick-austin, line 45: I don't think this is working for me. Below 832px (window width) the width scales with the window as I'd expect, and above 864px the width of the component is fixed to 800px. But for windows between 832px and 864px the right edge of the component is cropped off. I think it would just be a case of increasing `theme.spacing(4)` to `theme.spacing(8)`, presumably it needs to account for the padding of the root element and the paper element?

Screen Shot 2021-03-01 at 17 04 42

src/adminPage/adminPage.component.tsx
```diff
@@ -0,0 +1,232 @@
...
+    ScheduledMaintenanceState
+  >(props.scheduledMaintenance);
+  const [maintenance, setMaintenance] = useState<MaintenanceState>(
+    props.maintenance
+  );
+
+  return (
+    <div className={props.classes.root}>
+      <Typography variant="h3" className={props.classes.titleText}>
+        {getString(props.res, 'title')}
```
patrick-austin, line 93: I don't think it's necessarily a problem, but on wide windows the title being left aligned and the actual content being centred looks a bit strange to me. I know the title is left aligned on the cookies page, but on the login page (which also uses centred content) we don't have a left aligned `h3` header at the top of the page, just the `h5` header centred within the paper. I wonder if that might look better on the admin page?

Screen Shot 2021-03-01 at 17 10 59

patrick-austin (dismissed):
When navigating directly to the `/admin` url as a non-admin user, I get `Unhandled Rejection (Error): still loading auth provider`.

I didn't test the point about tampering with the token locally since I was getting this error anyway, and I wasn't sure what the best way to alter it was. How would you recommend going about that?

VKTB:
> When navigating directly to the `/admin` url as a non-admin user, I get `Unhandled Rejection (Error): still loading auth provider`.
I did not test this case but I was able to replicate this issue. I had a look and the reason why this is happening is because at the time of verifying the token during the mounting of the component, the provider is still an instance of `LoadingAuthProvider` and if you look at it, its `verifyLogIn` is set to reject requests hence why the error was thrown. 

I just noticed that the same error is in fact being thrown when I tamper the token in the local storage and navigate to `http://localhost:3000/admin` but the only difference is that I get logged out and the login page is displayed in this case. I think that the reason for this is because the application is reloaded when I navigate to `/admin` causing the site to be configured again, and because the token is tampered, the verification check that is being performed in the [configureSite](https://github.com/ral-facilities/scigateway/blob/542b6d165e3ab4e292add067eba7e5a77615376b/src/state/actions/scigateway.actions.tsx#L206) function fails and this is why I get logged out and the redirected to the login page.

Myself and Louise had a chat about this earlier today and we decided that the best place to do this check may be as part of the `componentDidUpdate` function when the `loading` props changes to false. However given that the application gets reloaded when navigating to /`admin` it means that the verification token check inside the `configureSite` function should prevent the user from accessing the website (and the admin page) if they tamper the token beforehand. Correct me if I am wrong @patrick-austin @louise-davies.

Another way that someone may be able to gain access to the admin page, which I have learnt during my current module, is by intercepting and modifying the responses sent from the API. For example, if someone was to intercept the token response that is sent back after a user logs in, they can modify the response so that `userIsAdmin` is `True`. This will trick the application into creating the admin route and displaying an _Admin_ menu item. The application does not get reloaded when a page is accessed by clicking on a button from a menu, therefore there won't be any checks that will stop the user from accessing and seeing the content in the admin page. Doing a verification token check as part of `componentDidUpdate` will not work because the loading of the site will already be completed, meaning that the `loading` prop will be `False` when the user navigates to the `Admin` page.
 
> I didn't test the point about tampering with the token locally since I was getting this error anyway, and I wasn't sure what the best way to alter it was. How would you recommend going about that?
Sorry I should have included some instruction but this is how I do it on Chrome:
1. Access the Dev Tools (F12)
2. Click on the _Application_ tab
3. Expand _Local Storage_ from the menu on the left and select _http://localhost:3000_
4. Locate the _scigateway:token_ key, double click on its value and paste your JWT token

louise-davies:
@VKTB So if I understand correctly - you're concerned that the token may be modified at any time, not just re: loading. So maybe the admin page needs to check whenever it renders instead of when it mounts/it's props update (aka whenever the user navigates to the `/admin` page) - so the check could be instead part of the render method instead?

src/adminPage/adminPage.component.tsx
```diff
@@ -0,0 +1,232 @@
...
+            setMaintenance({
+              ...maintenance,
+              message: e.currentTarget.value,
+            })
+          }
+        />
+        <div style={{ display: 'row' }}>
+          <FormControlLabel
+            style={{ float: 'left' }}
+            value={scheduledMaintenance.show}
```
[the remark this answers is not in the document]
VKTB: Oops, good spot, thanks!

src/adminPage/adminPage.component.tsx
VKTB: Thanks, that works!

src/adminPage/adminPage.component.tsx
VKTB: I kept it left aligned so that it's consistent with the _Contact_, _Help_ and _Cookies_ pages. I will be adding new content to the Admin page in another issue so I will decide what's best to do with the title after making the changes in that issue.

VKTB:
@VKTB Correct - the token can be modified by modifying what's in the storage or by intercepting and modifying the response to the login request. In my previous post, I was trying to say that if you modify the token in the local storage and navigate to `http://localhost:3000/admin` or any other protected route, then it won't be the verify check in `componentDidUpdate` that will log you out and prevent you from accessing the page but the check in the [configureSite](https://github.com/ral-facilities/scigateway/blob/542b6d165e3ab4e292add067eba7e5a77615376b/src/state/actions/scigateway.actions.tsx#L206) function instead. This is because the site gets re-configured when it is reloaded so this code is the first to get executed, rendering the check in the `componentDidUpdate` redundant. However, there is probably no harm in leaving it in as it could defend against other ways of modifying tokens of which I am not aware of at the moment. Just let me know what you think and I will either remove it or write a unit test and keep it in.

src/authentication/testAuthProvider.tsx
```diff
@@ -36,7 +40,7 @@ export default class TestAuthProvider implements AuthProvider {
   }
 
   public verifyLogIn(): Promise<void> {
-    return this.refresh();
+    return Promise.resolve();
```
louise-davies, line 43: Is there a reason for this change? This could cause unexpected behaviour in other tests.

VKTB: Oops, I was meant (but forgot) to revert this change. I can't remember why I actually made the change in the first place. Good spot, thanks!

louise-davies (approved):

PR #679, opened 2021-04-01
Display admin functionality from plugins

### LINKED ISSUE #625: Investigate how to display admin functionality from plugins
**Description**:
Plugins (e.g. datagateway-download) can have admin functionality, so we need a way for SciGateway to be able to display these and restrict access to only admin users, and for the plugins to be able to tell SciGateway which parts of them are admin-only.

**Acceptance criteria**:
- [ ] Plugins/parts of plugins can be loaded as admin-only and these are only accessibly by admins

VKTB:
## Description
Allows for DataGateway admin routes to be created if `"admin": true` is added to the route object in the `*-settings.json` file. Protects such routes so that only admins can access them by checking whether the user is admin or not and by not creating links to them in the navigation drawer. Displays the `PageNotFound` page if a non-admin user navigates to an admin route. This functionality is needed for the Admin Download table that will be created in DataGateway Download as part of the [Add TopCAT admin download table](https://github.com/ral-facilities/datagateway/issues/613) issue. In preparation for this work, I created two tabs in the Admin page - _Maintenance_ and _Admin Download_. The former is active by default and its panel displays the maintenance text boxes that admins can use to turn maintenance on and off, whereas the latter tab does not have a panel but instead is set to direct to the Admin Download table route `/admin-download` when an admin clicks on it.

Moved some of the admin plugin/ route checks to the `authorisedRoute` component as it makes more sense to have them in there though such checks are still required for a couple of things in the `routing` component - I left comments in the code that explains why they are needed. Also made some changes so that maintenance page is not displayed to admin users when they navigate to the plugins while the site is under maintenance.

While debugging some of the failing unit tests, I discovered that the state persists between unit tests. It turns out this is due to object mutation - when we are assigning values to the `state` object, we are passing the imported `initialState` constant to the `scigateway` object. This passes the memory pointer of `initialState` to `state.scigateway` so when we change any of the values of the `state.scigateway` in the tests, it also changes the values in `initialState`. The same thing happens with the `authorisation` property of the `initialState` constant because it is referencing the `authState` constant. To prevent these mutations, the `initialState` and the `authState` need to be spread. I applied these changes to the other unit tests in the code base as part of the [`f37e5c1`](https://github.com/ral-facilities/scigateway/pull/679/commits/f37e5c1fbb36bcde7c1cfeb27c497a794aaff2fa) commit.

I tried writing a unit test for the _Admin Download_ tab but I couldn't get it to work. Was testing for whether a `push('/admin-download')` action is sent to the store when this tab is clicked. I think it is to do with the click not simulating properly. Here is the code I wrote:
```TypeScript
it('redirects to Admin Download table when Admin Downlod tab is clicked', () => {
  const testStore = mockStore(state);
  const wrapper = mount(
    <Provider store={testStore}>
      <MuiThemeProvider theme={theme}>
        <MemoryRouter initialEntries={[{ key: 'testKey' }]}>
          <AdminPage />
        </MemoryRouter>
      </MuiThemeProvider>
    </Provider>
  );

  wrapper.find('#download-tab').hostNodes().simulate('click');

  expect(testStore.getActions().length).toEqual(1);
  expect(testStore.getActions()[0]).toEqual(push('/admin-download'));
});
```

## Testing instructions
Create an admin DataGateway Download route and treat that as the Admin Download table. This can be done by adding a route to the `datagateway-download-settings.json` file (note `"admin": true`):
```JSON
...
  "routes": [
    {
      "section": "Test",
      "link": "/admin-download",
      "displayName": "Admin Download",
      "admin": true,
      "order": 0
    },
...
  ],
...
```
Deploy SciGateway together with DataGateway Download.

- [ ] Review code
- [ ] Check Actions build
- [ ] The navigation drawer does not contain a link to the admin route.
- [ ] PageNotFound page is displayed when a non-admin user manually navigates to `/admin-download`.
- [ ] DataGateway Download table is displayed when an admin user manually navigates to `/admin-download`.
- [ ] _Maintenance_ tab is active by default in Admin page.
- [ ] An admin user is redirected to the DataGateway Download table after clicking on the _Admin Download_ tab from the Admin page.
- [ ] Create another non-admin route by either deploying another DataGateway plugin or adding a second route to the  `datagateway-download-settings.json` file. Make sure that a link is created for it in the navigation drawer and that both admin and non-admin users can access it.
- [ ] Put site into maintenance and check that the maintenance page is still displayed when a non-admin user navigates to a plugin but it is not displayed for non-admin users.

## Agile board tracking
closes #625

### DIFF
--- a/src/routing/authorisedRoute.component.test.tsx
+++ b/src/routing/authorisedRoute.component.test.tsx
@@ -51,11 +51,11 @@ describe('AuthorisedRoute component', () => {
   });
 
   it('renders non admin component when non admin user accesses it', () => {
+    const testAuthProvider = new TestAuthProvider('test-token');
+    testAuthProvider.isAdmin = jest.fn().mockImplementationOnce(() => false);
+    state.scigateway.authorisation.provider = testAuthProvider;
     state.scigateway.siteLoading = false;
     state.scigateway.authorisation.loading = false;
-    state.scigateway.authorisation.provider = new TestAuthProvider(
-      'test-token'
-    );
 
     const AuthorisedComponent = withAuth(ComponentToProtect);
     const wrapper = shallow(
--- a/src/routing/authorisedRoute.component.test.tsx
+++ b/src/routing/authorisedRoute.component.test.tsx
@@ -3,7 +3,7 @@ import withAuth from './authorisedRoute.component';
 import { createShallow } from '@material-ui/core/test-utils';
 import configureStore from 'redux-mock-store';
 import { StateType } from '../state/state.types';
-import { initialState } from '../state/reducers/scigateway.reducer';
+import { authState, initialState } from '../state/reducers/scigateway.reducer';
 import { createLocation } from 'history';
 import TestAuthProvider from '../authentication/testAuthProvider';
 import LoadingAuthProvider from '../authentication/loadingAuthProvider';
@@ -25,7 +25,7 @@ describe('AuthorisedRoute component', () => {
     shallow = createShallow({ untilSelector: 'div' });
 
     state = {
-      scigateway: initialState,
+      scigateway: { ...initialState, authorisation: { ...authState } },
       router: {
         action: 'POP',
         location: createLocation('/'),
@@ -35,7 +35,7 @@ describe('AuthorisedRoute component', () => {
     mockStore = configureStore();
   });
 
-  it('renders component when user logged in', () => {
+  it('renders non admin component when admin user accesses it', () => {
     state.scigateway.siteLoading = false;
     state.scigateway.authorisation.loading = false;
     state.scigateway.authorisation.provider = new TestAuthProvider(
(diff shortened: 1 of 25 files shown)


src/routing/authorisedRoute.component.test.tsx
```diff
@@ -35,15 +35,62 @@ describe('AuthorisedRoute component', () => {
...
     const AuthorisedComponent = withAuth(ComponentToProtect);
-    const wrapper = shallow(<AuthorisedComponent store={mockStore(state)} />);
+    const wrapper = shallow(
+      <AuthorisedComponent adminPlugin={false} store={mockStore(state)} />
+    );
+
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders non admin component when non admin user accesses it', () => {
```
sam-glendenning, line 53: This test and the one above it have the same content. Is the `isAdmin` function supposed to be mocked in each to return a different boolean value from the other? How are the tests knowing the user is an admin or not? Or have I got my wires crossed?

src/routing/authorisedRoute.component.test.tsx
```diff
@@ -35,15 +35,62 @@ describe('AuthorisedRoute component', () => {
...
+
+    const AuthorisedComponent = withAuth(ComponentToProtect);
+    const wrapper = shallow(
+      <AuthorisedComponent adminPlugin={false} store={mockStore(state)} />
+    );
+
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders admin component when admin user accesses it', () => {
```
sam-glendenning, line 68: Similarly here, how does the test know the user is an admin?

src/routing/authorisedRoute.component.test.tsx
VKTB: Good spot, thanks! Must've forgotten to change bits after pasting it.

sam-glendenning (approved):
All looks good to me.

agbeltran:
@louise-davies @VKTB to decide on URL of the admin download table plugin before merging.

VKTB:
> @louise-davies @VKTB to decide on URL of the admin download table plugin before merging.

I totally forgot to remind Louise that we need to decide on this. I think that it might be be better for the URL to be `/admin/download` or `/download/admin` instead of `/admin-download`. If you do decide to change it, please make sure that all `/admin-download` URL references in this and the [Display admin functionality from plugins](https://github.com/ral-facilities/scigateway/pull/679) PR are updated.

louise-davies (approved):
Tested this with the Admin Download table PR and besides a small dark mode issue it works fine  plus Sam also reviewed so I'll merge this in

PR #704, opened 2021-04-16
Redirect to plugin homepage #678

### LINKED ISSUE #678: Support redirection to plugin homepage
**Description**:
Depending on the plugin included, SciGateway should redirect to a specific plugin homepage. 

**Acceptance criteria**:
- [ ] SciGateway redirects to the homepage of a specific plugin

sam-glendenning:
## Description
Adding support for a new universal homepage to replace the basic SciGateway page if one is set. This is configured by a new field in public/settings.json:

`"homepageUrl": "/path/to/homepage"`

The homepage can be any registered route setup by a plugin in its own settings file, e.g. public/datagateway-dataview-settings.json. Ensure the "homepageUrl" in settings.json matches the path to this route exactly.

On website load, the new homepage is displayed first (also overriding any "startUrl" currently configured. This may be changed in a later commit). Any attempts to navigate to /, such as clicking on the SciGateway logo in the main app bar or through redirects, will redirect to the new homepage.

## Testing instructions
Best test case right now is by using the DataGateway homepage. This is currently in development so should be tested using the [customisable-homepage-#431](https://github.com/ral-facilities/datagateway/tree/customisable-homepage-%23431) branch on DataGateway. 
The homepage is currently stored in datagateway-common and imported into datagateway-dataview. So the minimum required plugin to load is datagateway-dataview. [Follow the instructions on the related PR to get started.](https://github.com/ral-facilities/datagateway/pull/662)

- [x] Review code
- [x] Check Actions build
- [x] Check that the new homepage loads when the website is first navigated to
- [x] Check that navigating to / in the address bar redirects to the new homepage
- [ ] Check that clicking the SciGateway logo on the main app bar redirects to the new homepage
- [x] Check that the new homepage is navigated to after logging in
- [x] Check that setting any other valid registered route as the new homepage passes the same tests as described above
- [x] Check that setting an invalid route as the new homepage displays a 404 page
- [x] Check that a link to the homepage does NOT display in the navigation drawer

## Agile board tracking
connect to [DataGateway PR #662](https://github.com/ral-facilities/datagateway/pull/662) and [DataGateway Issue #431](https://github.com/ral-facilities/datagateway/issues/431)
closes #678

### DIFF
--- a/src/routing/routing.component.tsx
+++ b/src/routing/routing.component.tsx
@@ -6,7 +6,7 @@ import {
   StyleRules,
   WithStyles,
 } from '@material-ui/core/styles';
-import { Route, Switch } from 'react-router';
+import { Route, Switch, Redirect } from 'react-router';
 import { StateType } from '../state/state.types';
 import { MaintenanceState, PluginConfig } from '../state/scigateway.types';
 import { connect } from 'react-redux';
@@ -76,7 +76,14 @@ class Routing extends React.Component<
         })}
       >
         <Switch>
-          <Route exact path="/" component={HomePage} />
+          {/* Redirect to a homepageUrl if set. Otherwise, route to / */}
+          {this.props.homepageUrl &&
+          this.props.location === '/' &&
+          this.props.homepageUrl !== '/' ? (
+            <Redirect to={this.props.homepageUrl} />
+          ) : (
+            <Route exact path="/" component={HomePage} />
+          )}
           <Route exact path="/contact" component={ContactPage} />
           <Route exact path="/help" component={HelpPage} />
           {this.props.userIsloggedIn && this.props.userIsAdmin ? (
--- a/src/routing/routing.component.tsx
+++ b/src/routing/routing.component.tsx
@@ -76,34 +76,34 @@ class Routing extends React.Component<
         })}
       >
         {/* Redirect to a homepageUrl if set. Otherwise, route to / */}
-        {this.props.homepageUrl &&
-        this.props.location === '/' &&
-        this.props.homepageUrl !== '/' ? (
-          <Redirect to={this.props.homepageUrl} />
-        ) : (
-          <Switch>
-            <Route exact path="/" component={HomePage} />
-            <Route exact path="/contact" component={ContactPage} />
-            <Route exact path="/help" component={HelpPage} />
-            {this.props.userIsloggedIn && this.props.userIsAdmin ? (
-              <Route exact path="/admin" component={withAuth(AdminPage)} />
-            ) : null}
-            <Route exact path="/login" component={LoginPage} />
-            <Route exact path="/cookies" component={CookiesPage} />
-            {this.props.maintenance.show ? (
-              <Route component={MaintenancePage} />
+        <Switch>
+          <Route exact path="/">
+            {this.props.homepageUrl && this.props.homepageUrl !== '/' ? (
+              <Redirect to={this.props.homepageUrl} />
             ) : (
-              this.props.plugins.map((p) => (
-                <Route
-                  key={`${p.section}_${p.link}`}
-                  path={p.link}
-                  render={() => <AuthorisedPlugin id={p.plugin} />}
-                />
-              ))
+              <HomePage />
             )}
-            <Route component={withAuth(PageNotFound)} />
-          </Switch>
-        )}
+          </Route>
+          <Route exact path="/contact" component={ContactPage} />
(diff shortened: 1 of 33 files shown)


sam-glendenning:
There's a few things to clarify with this PR before it can get to the stage of a full review. This will help us flesh out the system for new plugins down the line.

### 404 page if "homepageUrl" does not point to a valid registered route
If the "homepageUrl" in settings.json does not point to a registered route, a 404 page will display every time the user tries to navigate to the homepage. This isn't ideal and at the moment it results in the 404 page actually being the first thing the user sees when they visit the website due to the way that "homepageUrl" works very similarly to "startUrl". There are two ways to deal with this: either we leave this as it is to ensure that the "homepageUrl" is also correct or, my preferred option, have some sort of error checking similar to "startUrl" whereby the original SciGateway homepage is displayed instead.

### Do we still want to keep "startUrl"?
At the moment they function very similar and, if both are present in settings.json, the homepage is displayed instead, rendering the "startUrl" useless if a homepage is specified. Do we still want to keep "startUrl" and have it work in tandem with "homepageUrl" (albeit differently than it currently is) or shall we remove it altogether?

### SciGateway logo is still present in main app bar
This PR has made no changes to the logo present at the top of the page or the title of the website. With a new homepage now being possible, many users could be unaware of SciGateway at all and simply know the new homepage to be the old homepage. So the SciGateway logo (redirecting to DataGateway) and the SciGateway page title could be a bit confusing. Shall we replace them? If so, with what?

src/state/middleware/scigateway.middleware.tsx
```diff
@@ -103,6 +103,17 @@ export const listenToPlugins = (
                 scigateway: { startUrl: pluginMessage.detail.payload.link },
               })
             );
+          } else if (
```
sam-glendenning, line 106: Will be refactored if we decide "homepageUrl" will behave differently to "startUrl". Besides, this code only ensures that the URL is navigated to when the app is first launched and doesn't handle any future redirects.

sam-glendenning:
Current state: writing tests for new code and looking for thoughts on code and structure so far

sam-glendenning:
Not sure if I can do Cypress tests for this... I think you'll have to rely on manual testing and unit testing to be sure you're getting the result you want.

- if `"homepageUrl"` is set to a valid route in `settings.json`, any navigations to `/` should redirect to the homepageUrl
- if `"homepageUrl"` is not in `settings.json` at all, any navigations to `/` do not redirect and the _SciGateway_ homepage is displayed
- if `"homepageUrl"` is set to an _invalid_ route in `settings.json`, any navigations to `/` redirect to `/login` if the user is not logged in and load the 404 component if the user is logged in. If we don't want this behaviour, I can try and make a workaround for it

sam-glendenning:
The "startUrl" is potentially being rendered useless by this PR. We should consider removing it but see #708 for more info.

sam-glendenning:
@VKTB @louise-davies since I can't officially re-request a review, this message lets you know it's ready for a proper review now

sam-glendenning:
@louise-davies this PR along with [the matching one on DataGateway](https://github.com/ral-facilities/datagateway/pull/662) is now ready for review. 

Removing Viktor as an assigned reviewer as he is leaving SEG.

src/routing/routing.component.tsx
```diff
@@ -74,28 +75,35 @@ class Routing extends React.Component<
...
-          <Route component={withAuth(PageNotFound)} />
-        </Switch>
+        {/* Redirect to a homepageUrl if set. Otherwise, route to / */}
+        {this.props.homepageUrl &&
+        this.props.location === '/' &&
+        this.props.homepageUrl !== '/' ? (
+          <Redirect to={this.props.homepageUrl} />
+        ) : (
+          <Switch>
+            <Route exact path="/" component={HomePage} />
```
louise-davies, line 85: Can this be refactored to have one "HomePage" route like this:

```
<Route exact path="/">
  {this.props.homepageUrl && this.props.homepageUrl !== '/'  ? <Redirect to={this.props.homepageUrl} /> : <HomePage/>}
</Route>
```

sam-glendenning: I see what you did there. Much cleaner

louise-davies: https://reactrouter.com/web/api/Redirect - our use case is quite similar to the react-router example!

louise-davies (approved):

PR #736, opened 2021-06-02
Bugfix/fix plugin loading #732

### LINKED ISSUE #500: darkMode local storage always false in Travis test
**Description**:
Unit test 'should load dark mode preference into store' passes locally, but in Travis it has started failing. Logging seems to indicate that the local darkmode setting is always false, so the test fails. This started to happen after cardView was merged, but not immediately after so may be unrelated?

Still failing with the change to GHA

Marking test to be skipped as part of PR #443 

**Acceptance criteria**:
- [ ] Get the test passing in GitHub actions again

### LINKED ISSUE #732: Issue when reloading page
**Description**:
The first time the site loads, all works ok. But when reloading the same page, an error occurs "datagateway-dataview failed to load correctly". Racing condition between scigateway and plugins?

**Acceptance criteria**:
- [ ] Fix issue when reloading page.

louise-davies:
## Description
Fixes the bug where plugins would sometimes not be loaded when loaded directly (e.g. when trying to load `/browse/investigation`). This bug was discovered due to the changes in dataview to allow plugin routes to be configurable, which meant they were loaded later on after the plugin settings files were loaded which then revealed this issue. This PR fixes the issue by checking if we are on a "potential plugin route" (i.e. any route that scigateway doesn't know about) and listening for the register route event from the plugin that corresponds to our current location. I also added a 3 second timeout to this listening, as otherwise the site would never load if someone navigated to a route that should 404.

Additionally, since I was looking at `scigateway.actions.test.tsx`, I had a look at the darkMode test which was skipped and realised it wasn't mocking localStorage. Mocking it makes the test pass in CI again so this PR also fixes #500 

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Check that when navigating directly to a plugin route the plugin always loads correctly. I will shortly be adding this to diamond preprod so this can be checked there (and behaviour contrasted against any of the other machines)
 
## Agile board tracking
Fixes #732

### DIFF
--- a/src/state/actions/scigateway.actions.test.tsx
+++ b/src/state/actions/scigateway.actions.test.tsx
@@ -69,6 +69,11 @@ describe('scigateway actions', () => {
       .mockImplementation(() => Promise.resolve());
   });
 
+  afterEach(() => {
+    jest.runOnlyPendingTimers();
+    jest.useRealTimers();
+  });
+
   it('toggleDrawer only needs a type', () => {
     const action = toggleDrawer();
     expect(action.type).toEqual(ToggleDrawerType);
--- a/src/state/actions/scigateway.actions.test.tsx
+++ b/src/state/actions/scigateway.actions.test.tsx
@@ -38,6 +38,7 @@ import { StateType } from '../state.types';
 import loadMicroFrontends from './loadMicroFrontends';
 import log from 'loglevel';
 import { createLocation } from 'history';
+import { flushPromises } from '../../setupTests';
 
 function mockAxiosGetResponse(message: string): void {
   (mockAxios.get as jest.Mock).mockImplementationOnce(() =>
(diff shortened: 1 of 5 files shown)


src/state/actions/scigateway.actions.test.tsx
```diff
@@ -52,6 +52,8 @@ function mockAxiosGetResponse(message: string): void {
 
 describe('scigateway actions', () => {
   beforeEach(() => {
+    jest.useFakeTimers();
```
sam-glendenning, line 55: Worth adding this?

```
afterEach(() => {
    jest.runOnlyPendingTimers();
    jest.useRealTimers();
});
```

Good practice according to https://testing-library.com/docs/using-fake-timers/

sam-glendenning (dismissed):
Great fix! Tested on a variety of routes on each plugin and never saw the error message. Only downside is the slightly longer loading times due to the 3 second delay but I don't think there's much we can do about that.

src/state/actions/scigateway.actions.test.tsx
louise-davies: Good spot - I'll add this in and add it to other test files that also use fake timers 👍

louise-davies:
> Great fix! Tested on a variety of routes on each plugin and never saw the error message. Only downside is the slightly longer loading times due to the 3 second delay but I don't think there's much we can do about that.

Technically the 3 second delay only occurs if either a user is navigating to a page that would end up in a 404 (which hopefully should be quite rare!) or if the plugins take longer than 3 seconds to send the relevant register route method.

sam-glendenning (approved):

PR #740, opened 2021-06-17
Fixing login token bugs #735

### LINKED ISSUE #735: Autologin session id is null after you visit login page
**Description**:
If you go to the login page and then navigate away (i.e user considers loggin in then continues to browse anonymously), the user is still logged in but if the go to dataview they get errors saying their session id is null. Visiting the login page must be messing with the login token somehow.

**Acceptance criteria**:
- [ ] User can visit login page and navigate away and continue to browse data

sam-glendenning:
## Description
This fixes issues which, through certain scenarios, would cause the user to be logged out of SciGateway when they shouldn't be. What would happen is the user's login session would be unexpectedly invalidated (their token in localStorage would be removed) which would cause errors when they try to navigate plugins. This was specifically a problem for the anon authenticator, which assigns a token automatically on initial page load; this bug would cause the token to be invalidated and errors would occur when the user navigated to plugins.

The issue was determined to be caused by two things:
1. when navigating to the login page and changing the authenticator selection from the dropdown but not actually logging in
2. when a login attempt failed

These changes now prevent this from happening and ensure that the autoLogin anon session remains valid. A further change has also been made which prevents a logged-in user from navigating to `/login` at all (previously the user could access it through direct navigation via the address bar, for example). Now what happens is the logged-in user will be redirected from `/login` to the homepage. The only way to access /login again is to explicitly log out (or be explicitly logged out, e.g. token expiry). The user can tell if they are logged out by looking for the 'Sign in' button at the top right of the page. Overall, this brings SciGateway in line with how other major websites typically handle login sessions - you are not allowed to login or attempt to login to an account while you are already logged in.

## Testing instructions
Load SciGateway with at least one plugin, ideally Dataview. Test with different authenticators, I used https://datagateway-dev.isis.stfc.ac.uk/api for my `authUrl` as it provides multiple authenticators. Keep an eye on the site's localStorage key-value pairs so you can see exactly when `scigateway:token` is being added, changed and removed. Check a token is assigned on initial page load and that it stays there while navigating different pages. Try changing the login authenticators and signing in using invalid credentials. The token should still be there. Also try logging in successfully to make sure the token changes. Do the same tests with this token to ensure it doesn't get removed unexpectedly. Hopefully you shouldn't see any errors when navigating the site!

Also try to navigate to `/login` while logged in. You should be redirected to the homepage every time. Then logout and make sure you can access `/login` again.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #735

### DIFF
--- a/src/loginPage/loginPage.component.tsx
+++ b/src/loginPage/loginPage.component.tsx
@@ -305,6 +305,10 @@ const LoginPageComponent = (props: CombinedLoginProps): React.ReactElement => {
   );
 
   React.useEffect(() => {
+    setMnemonic(props.auth.provider.mnemonic);
+  }, [props.auth.provider.mnemonic]);
+
+  React.useEffect(() => {
     if (typeof mnemonic !== 'undefined' && !fetchedMnemonics) {
       fetchMnemonics(authUrl).then((mnemonics) => {
         const nonAdminAuthenticators = mnemonics.filter(
--- a/src/loginPage/loginPage.component.tsx
+++ b/src/loginPage/loginPage.component.tsx
@@ -319,6 +319,10 @@ const LoginPageComponent = (props: CombinedLoginProps): React.ReactElement => {
   }, [mnemonic, fetchedMnemonics, authUrl]);
 
   React.useEffect(() => {
+    setMnemonic(props.auth.provider.mnemonic);
+  }, [props.auth.provider.mnemonic]);
+
+  React.useEffect(() => {
     if (
       props.auth.provider.redirectUrl &&
       props.location.search &&
--- a/src/loginPage/loginPage.component.tsx
+++ b/src/loginPage/loginPage.component.tsx
@@ -319,7 +319,12 @@ const LoginPageComponent = (props: CombinedLoginProps): React.ReactElement => {
   }, [mnemonic, fetchedMnemonics, authUrl]);
 
   React.useEffect(() => {
-    setMnemonic(props.auth.provider.mnemonic);
+    if (
+      typeof props.auth.provider.mnemonic !== 'undefined' &&
+      props.auth.provider.mnemonic !== ''
+    ) {
+      setMnemonic(props.auth.provider.mnemonic);
+    }
   }, [props.auth.provider.mnemonic]);
 
   React.useEffect(() => {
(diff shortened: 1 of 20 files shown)


sam-glendenning:
Current progress on this:
- dropdown is not displaying when navigating directly to /login
- snapshots are incorrectly passing because they have been updated wrongly (likely a consequence of the above). See loginPage.component.test lines 199 and 231
- coverage is way down but this will be fixed after the above issues

sam-glendenning:
The following changes have now been made:

1. - the login page is now inaccessible while the user is logged in. They must be explicitly logged out to navigate to this page. Any attempt to do so will `<Redirect>` them to the homepage
2. - because of this, earlier commits that made changes that kept the user logged in if they tried to login to another account while logged in have been invalidated. The user is now unable to do this so this code is not necessary. This is probably only relevant to Goel as he is the only one I know of who has seen this functionality. 

@GoelBiju @louise-davies this is now ready for a second review. New changes have been made to the routing component, make sure to double check its new snapshots are correct.

src/loginPage/loginPage.component.tsx
```diff
@@ -88,9 +85,10 @@ interface LoginPageProps {
 interface LoginPageDispatchProps {
   verifyUsernameAndPassword: (
     username: string,
-    password: string
+    password: string,
+    mnemonic: string | undefined,
+    authUrl: string | undefined
```
louise-davies, line 90: Small style comment, but I prefer the syntax `mnemonic?: string` rather than explicitly saying `mnemonic: string | undefined`

src/loginPage/loginPage.component.tsx
```diff
@@ -226,13 +240,12 @@ export const AnonLoginScreen = (
 );
 
 export const LoginSelector = (
-  props: CombinedLoginProps & { mnemonics: ICATAuthenticator[] }
+  props: CombinedLoginProps & {
+    mnemonics: ICATAuthenticator[];
+    mnemonic: string | undefined;
+    setMnemonic: React.Dispatch<React.SetStateAction<string | undefined>>;
```
louise-davies, line 246: Another small style comment. This feels overly complex - I'm sure the type `(mnemonic: string | undefined) => void` is probably fine

louise-davies (approved):

PR #766, opened 2021-09-06
Removing `startUrl` functionality and tests #708

### LINKED ISSUE #708: Remove "startUrl" from settings.json
**Description**:
I decided to make this a separate issue instead of making it part of the PR to add a plugin homepage.

With the addition of a [redirect to an optional plugin homepage](https://github.com/ral-facilities/scigateway/pull/704), the `"startUrl"` field in `settings.json` is potentially being rendered useless. The only time it may be relevant is if a new homepage is not set with `"homepageUrl"`, in which case it continues to function as normal. In case people aren't aware of the exact differences between the two:

- `"startUrl"` redirects to a specified URL when the user navigates to the website root
- `"homepageUrl"` redirects to a specified URL whenever the user navigates to the website root and when the homepage logo at the top of the screen is clicked

Therefore, when a `"homepageUrl"` is set, the `"startUrl"` is never used. I'm trying to think of potential use cases where SciGateway may still want `"startUrl"` and not `"homepageUrl"` and there aren't enough to justify its inclusion in my opinion. So we should work to remove it.

**Acceptance criteria**:
- [x] Delete all references to startUrl in code
- [x] Check required functionality can still be achieved with `"homepageUrl"` alone

sam-glendenning:
## Description
This removes all functionality relating to the `startUrl` setting found in `settings.json`. None of this is required anymore due to the work done in https://github.com/ral-facilities/scigateway/pull/704 which redirects all traffic to a custom plugin homepage if set. `startUrl` was always used to redirect traffic that went directly to the website root to a set plugin path, e.g. `/browse/investigation`. However, this isn't required anymore because, in production, a custom homepage will always be set that we'll want to redirect to instead (or, failing that, the SciGateway homepage).

This PR removes all references to `startUrl` as well as any tests related to it. Also changing a couple of names in code to `homepageUrl` where relevant functionality overlaps (`homepageUrl` is the new setting that has replaced `startUrl`, this functionality was merged in the above mentioned pull request).

## Testing instructions
Check that redirects work as expected and any `startUrl` setting mentioned in `settings.json` doesn't do anything anymore. Add a `homepageUrl` to check that the homepage functionality continues to work as expected, e.g.:
```
"homepageUrl": "/datagateway"
```

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes #708

### DIFF
--- a/src/routing/__snapshots__/authorisedRoute.component.test.tsx.snap
+++ b/src/routing/__snapshots__/authorisedRoute.component.test.tsx.snap
@@ -78,7 +78,7 @@ exports[`AuthorisedRoute component renders non admin component when non admin us
 </div>
 `;
 
-exports[`AuthorisedRoute component renders redirect when startUrl is configured and logged in 1`] = `
+exports[`AuthorisedRoute component renders redirect when homepageUrl is configured and logged in 1`] = `
 <div>
   <Redirect
     push={true}
--- a/src/routing/authorisedRoute.component.test.tsx
+++ b/src/routing/authorisedRoute.component.test.tsx
@@ -95,13 +95,13 @@ describe('AuthorisedRoute component', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
-  it('renders redirect when startUrl is configured and logged in', () => {
+  it('renders redirect when homepageUrl is configured and logged in', () => {
     state.scigateway.siteLoading = false;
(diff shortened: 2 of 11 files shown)


GoelBiju (approved):

agbeltran:
thanks @sam-glendenning !

PR #767, opened 2021-09-07
Added the alt text to homepage images from the default.json file #753

### LINKED ISSUE #753: Alternate Text for Homepage Images
**Description**:
On the homepage, the three images in the 'How it works...' section and the data centre banner need alternative text for them (circled in the screenshot). The three circle images have the `alt` attribute defined but it is currently blank. 
image

The big SciGateway logo (on top of the banner) already has alt text. [This is fetched from the language file](https://github.com/ral-facilities/scigateway/blob/master/src/homePage/homePage.component.tsx#L90), so we probably want to take a similar approach with these other four images.

While this is configurable, it's probably easier to decide on the text for our implementation before doing the issue (hence the question label)?

For whoever comes to do this, in the code, the images imported as `Image` (data centre banner), `ExploreImage`, `AnalyseImage` and `RecordImage`.

**Acceptance criteria**:
- [x] The four images listed above should have an `alt` attribute, describing what the picture contains
- [x] When using a screen reader, it should tell the user what the image is (i.e. text-to-speech on the alt text

AkhilMag13:
## Description
Added the alt text for the images and it is read out correctly by the screen reader.
## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #753

### DIFF
--- a/public/res/default.json
+++ b/public/res/default.json
@@ -19,7 +19,9 @@
   },
   "home-page": {
     "title": "SciGateway",
+    "logo-label": "SciGateway Logo",
     "how-label": "How it works...",
+    "background-label": "NERC JASMIN supercomputer based at STFC’s Rutherford Appleton Laboratory",
     "explore-label": "1. Explore",
     "explore-description": "Explore and visualise your data. Select and pre-process data to pull in to your analysis environment.",
     "analyse-label": "2. Analyse",
--- a/public/res/default.json
+++ b/public/res/default.json
@@ -21,7 +21,6 @@
     "title": "SciGateway",
     "logo-label": "SciGateway Logo",
     "how-label": "How it works...",
-    "background-label": "NERC JASMIN supercomputer based at STFC’s Rutherford Appleton Laboratory",
     "explore-label": "1. Explore",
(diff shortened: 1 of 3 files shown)


public/res/default.json
```diff
@@ -19,7 +19,9 @@
   },
   "home-page": {
     "title": "SciGateway",
+    "logo-label": "SciGateway Logo",
     "how-label": "How it works...",
+    "background-label": "NERC JASMIN supercomputer based at STFC’s Rutherford Appleton Laboratory",
```
sam-glendenning, line 24: I remember we agreed that this would be good alt text for the background image behind the SciGateway logo but it is not being used on the image as it stands. Personally I am fine with this, because it could conflict with the alt text on the logo itself but if we choose to leave it off the image, we should remove this from `default.json`

sam-glendenning (changes requested):

public/res/default.json
AkhilMag13: Yes, I thought this too, but left it in as it might be useful in the future if needed. I have removed it now and committed as it is cleaner.

sam-glendenning (approved):

PR #800, opened 2021-11-05
#799 - change id of base div from root to scigateway

### LINKED ISSUE #799: Modify index.html to have a tag/some text that icinga can monitor
**Description**:
Currently, the icinga check for SciGateway checks that `<title>SciGateway</title>` exists in `index.html` - however this means we are unable to change the page title without triggering icinga warnings. We need to modify the icinga check - but we need to edit `index.html` to have something icinga can search for to ensure the page is being served. I'd say the most appropriate place is to edit `<div id="root">` to be `<div id="scigateway">` as this won't affect how the app or the plugins work (I think!) and isn't anything user-facing and thus doesn't need to be configurable.

**Acceptance criteria**:
- [ ] `index.html` edited to have a non-configurable, uniquely identifying feature that can be asserted for in icinga
- [ ] Update the icinga check to search for the new feature

louise-davies:
## Description
This means we can have an icinga check based on this div (which is only used for telling React where to load, and nothing user facing) rather than the title (which is user facing and thus we'd like to be able to configure)

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #799

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -28,7 +28,7 @@
     <script crossorigin src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
     <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
     <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="root"></div>
+    <div id="scigateway"></div>
     <!--
       This HTML file is a template.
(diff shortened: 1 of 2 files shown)


joelvdavies (approved):
Checked changes, tests pass, runs locally.

PR #822, opened 2021-11-23
Broadcast signout action when user signs out

louise-davies:
## Description

This is so plugins can listen to this and perform actions such as cache invalidation.

## Testing instructions
If you want to test this, you need to login as two users with different data. I can show some gifs/do a demo of the different behaviours when using two users on DLS preprod data.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
This is required by ral-facilities/datagateway#928


joelvdavies (approved):
Reviewed changes, tests pass, checked action received in search and dataview. I will take your word for it that the issue is fixed since I cannot test it directly :).

PR #886, opened 2021-12-16
#860 Autologin on logout & remove anon authenticator

### LINKED ISSUE #860: Autologin when a user logs out and remove anon authenticator from login dropdown
**Description**:
We want to remove the anon authenticator from the login dropdown - however that means that a user would have no way to login as anonymous if they logged into their account and then logged out later, as the autoLogin code only runs upon page load. We should run the autoLogin code when a user logs out, and then it's safe to remove the anon authenticator option from the login dropdown.

**Acceptance criteria**:
- [ ] Anon authenticator does not appear in the login dropdown
- [ ] User can login to their account and logout again and still have access via an anon user session

louise-davies:
## Description
Adds an autoLoginMiddleware that handles autoLogin when a user is logged out and also we filter out anon mnemonics from the login page

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #860


joelvdavies (approved):
Looks good. Reviewed code and coverage changes, CI passes, checked anon was filtered out.

PR #924, opened 2022-01-26
Refactor single-spa plugin mounting code

louise-davies:
## Description
The single-spa config now takes into account the routes that plugins say they should be on. This ensures that plugins actually unmount correctly and we don't have detached from the DOM but mounted in React components/plugins.

This is what I believe was causing the original duplicate queries bug, and some problems for @joshuadkitenge on his search tabs issue

Requires ral-facilities/datagateway#1089

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


louise-davies:
https://user-images.githubusercontent.com/22525471/151532489-5d6e7cc9-695a-4ed0-bb27-5478d8b62a7c.mp4

Here's a video with all the plugins working in SciGateway with the new loading method. I used a video rather than a GIF because it's quite long

joelvdavies (approved):
Reviewed code and coverage changes. Tested plugins behaved correctly. CI passes.

PR #929, opened 2022-02-03
Added loading indicator to show progress when delay in showing option

### LINKED ISSUE #928: Sign-in page - delay in showing authenticator options
**Description**:
Reported by Franz

  -	Sometimes the login page would only show the lock symbol and “Sign in” but not the actual fields for the credentials until quite a wait:
 
 
signin

**Acceptance criteria**:
- [ ] Check what is causing delays in showing authenticators and display a message if this happens

AkhilMag13:
## Description
Added a loading indicator to show progress while the authenticator options are not yet loaded.

## Testing instructions
Throttling the internet connection to slow 3G to see effect
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #928


joshuadkitenge:
login loading indicator

src/loginPage/loginPage.component.tsx
```diff
@@ -416,6 +426,7 @@ const LoginPageComponent = (props: CombinedLoginProps): React.ReactElement => {
   const authUrl = props.auth.provider.authUrl;
   const [mnemonics, setMnemonics] = useState<ICATAuthenticator[]>([]);
   const [fetchedMnemonics, setFetchedMnemonics] = useState<boolean>(false);
+  const [isComponentVisible, setComponentVisible] = useState<boolean>(false);
```
louise-davies, line 429: Slight comment that could this be renamed to be something like `loading` and `setLoading` - you'd have to reverse the logic (or use `loaded` and `setLoaded` I suppose...) - `isComponentVisisble` is a bit confusing

louise-davies (changes requested):
Rather than pass `setComponentVisible` to a bunch of components that don't really need it (i.e. they don't have any sort of loading), I think we can instead remove it by having the loading component run on the same logic as what triggers the mnemonic fetching in the first place i.e

```
{typeof mnemonic !== 'undefined' && !fetchedMnemonics ? (
    <CircularProgress className={props.classes.spinner} />
) : null}
```

Also, we don't need to different `CircularProgress` components - we can just use or conditions to have it render when either `props.auth.loading || typeof mnemonic !== 'undefined' && !fetchedMnemonics`

src/loginPage/loginPage.component.tsx
```diff
@@ -551,6 +553,10 @@ const LoginPageComponent = (props: CombinedLoginProps): React.ReactElement => {
         {props.auth.loading ? (
           <CircularProgress className={props.classes.spinner} />
         ) : null}
+
+        {typeof mnemonic !== 'undefined' && !fetchedMnemonics ? (
+          <CircularProgress className={props.classes.spinner} />
+        ) : null}
```
louise-davies, line 559: These could still be incorporated into one conditional - I don't know if that's more or less readable

joshuadkitenge: It's still readable. I have added the changes

louise-davies (approved):

PR #967, opened 2022-02-24
Navigation drawer logo is defined by width instead of the height #964

### LINKED ISSUE #964: Nav bar logo size defined by height instead of width
**Description**:
Currently, the size of the navbar logo is constrained using `height` rather than `width`. Whilst using `height` displays the correct result for the default logo, if any logo has different dimensions then it will look off. The real size constraint for the logo is that it must fit within the sidebar, and thus it should be constrained by a `width` instead.

Example:

Current behaviour, `img` tag has `height: 40px` set:

image

Desired behaviour, `img` tag with `width: 188px` set:

image

**Acceptance criteria**:
- [ ] Logos with different dimensions are displayed correctly in the sidebar

joshuadkitenge:
## Description
The navigation drawer logo should be set by the width of the of navigation bar 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #964

### DIFF
--- a/src/navigationDrawer/navigationDrawer.component.tsx
+++ b/src/navigationDrawer/navigationDrawer.component.tsx
@@ -61,7 +61,7 @@ const styles = (theme: Theme): StyleRules =>
     menuLogo: {
       paddingRight: theme.spacing(2),
       paddingLeft: theme.spacing(2),
-      height: 40,
+      width: '188px',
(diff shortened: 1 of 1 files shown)


AkhilMag13 (approved):
Works as expected

PR #972, opened 2022-03-01
Improve tour

louise-davies:
## Description
Modifications to the tour component to work a bit better - one is that we no longer close the drawer, and the other is to prevent scrolling. There's an issue in the underlying library to be able to disable scrolling only for specific tour steps, which we should use when it's fixed & released.

Also, I added some QOL changes to the NodeJS servers

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Goes with ral-facilities/datagateway#1138


joelvdavies (approved):
Reviewed changes, CI passes, tour works as expected.

PR #974, opened 2022-03-03
Making default admin view configurable #958

### LINKED ISSUE #958: Make default admin view configurable
**Description**:
TopCAT currently displays the admin download table as the default view when looking at the admin page. We should try and make this work and make it configurable.

The first problem I see is that technically the admin download table isn't "in" the admin page, and instead when you click on the tab there's a link to it. We should see if we can actually host it in the page so that the tabs work as expected (as otherwise if admin download was the default you'd currently never see the messages page). Then we can add a config option to make a certain tab the default.

**Acceptance criteria**:
- [ ] Admin download page is actually hosted in the admin page
- [ ] Admin page default tab is configurable

joshuadkitenge:
## Description
- Rendering the download table plugin in div within the admin page which includes the Tabs
- allow this page to be configurable  using the settings file 

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #958


joshuadkitenge:
ral-facilities/datagateway/pull/1140   pr on datagateway

src/adminPage/adminPage.component.test.tsx
```diff
@@ -10,22 +10,26 @@ import { MuiThemeProvider } from '@material-ui/core';
...
+import { act } from 'react-dom/test-utils';
+
 describe('Admin page component', () => {
   let mount;
   let mockStore;
   let state: StateType;
+  let history: History;
+  const setStateMock = jest.fn();
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const useStateMock: any = (useState: any) => [useState, setStateMock];
```
louise-davies, line 25: I'm not a fan of mocking `useState` - from what I can tell it's used to assert that the tab change is done - can we not just test that we find what we expect according to the tab e.g. on the download page we expect a div with id `datagateway-download`, on the maintenance page you can check we have the textboxes or buttons etc.

louise-davies (changes requested):
The admin page seems to be working well, but for some reason the main download page is broken - when I click on the link in the sidebar I get a 404 page

Can this be investigated?

src/mainAppBar/mainAppBar.component.tsx
```diff
@@ -133,6 +134,12 @@ const MainAppBar = (props: CombinedMainAppBarProps): React.ReactElement => {
 
   const location = useLocation();
 
+  const navigateToAdminPage = (): void => {
+    props.navigateToAdminPage(
+      adminRoutes[props.adminPageDefaultTab ?? 'maintenance']
+    );
+  };
+
```
louise-davies, line 142: This could be moved so it's defined entirely in the `mapDispatchToProps` using the second argument that gets passed to `mapDispatchToProps`: `ownProps` to access the default admin tab

src/routing/routing.component.tsx
```diff
@@ -79,27 +79,23 @@ export const AuthorisedAdminPlugin = withAuth(true)(PluginPlaceHolder);
...
 
-const getPluginRoutes = (
-  plugins: PluginConfig[]
+export const getPluginRoutes = (
+  plugins: PluginConfig[],
+  admin?: boolean
 ): {
   [plugin: string]: string[];
 } => {
+  // console.log('POP RENDERS PAGE');
```
louise-davies, line 88: Remove

src/adminPage/adminPage.component.tsx
```diff
@@ -67,214 +62,113 @@ const styles = (theme: Theme): StyleRules =>
...
+            hidden={tabValue !== 'maintenance'}
+          >
+            <MaintenancePage />
           </div>
-        </Paper>
-      </div>
+        </Route>
+
+        {Object.entries(pluginRoutes).map(([key, value]) => {
+          const pluginId = key.endsWith('_admin') ? key.slice(0, -6) : key;
```
louise-davies, line 147: `getPluginRoutes` no longer appends `_admin` to admin routes so this logic could be simplified to just use `key`

joshuadkitenge:
@louise-davies. fixing some issue due to merging the latest master

joshuadkitenge:
> @louise-davies. fixing some issue due to merging the latest master

It's ready for review now. I didn't pull DataGateway correctly, that is what was causing my problems

joshuadkitenge:
Changed the assertions to check for download plugin id and the maintenance page id

louise-davies (approved):
Looks good and works well now 👍

joshuadkitenge:
admin page

PR #985, opened 2022-03-11
Fix plugin loading bug

louise-davies:
## Description
After a lot of debugging and page refreshing to try and trigger the bug with logging, I think it was caused by `singleSpa.start()` being called before plugins had a chance to send their routes and thus make single-spa aware of them. Fixed by moving the `singleSpa.start()` call in the attempt to wait for plugins to send their routes section of `configureSite`.

Also, if the plugins are really slow I've added in a `triggerAppChange` call in the middleware.

EDIT: After noticing the bug still sometimes occurred, and a bug in that when the admin download table was directly accessed it loaded infinitely, I have added a "general" fix i.e. a setInterval that basically polls until the plugin routes has loaded, and checks if the preloader exists and if it does it calls `singleSpa.unloadApplication` (which forces a reload). I'd expect that this should cover all other cases of the bug happening but let me know if you still see it!

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] I've put these changes on DG ISIS prod, but didn't find a way to reliably recreate the bug. I basically just refreshed the page on my phone a bunch until the preloader takes a while to disappear. So either trust me or have fun refreshing the page 😅

### DIFF
--- a/src/state/actions/scigateway.actions.test.tsx
+++ b/src/state/actions/scigateway.actions.test.tsx
@@ -580,9 +580,9 @@ describe('scigateway actions', () => {
       },
     });
 
-    await asyncAction(dispatch, getState);
-
     const eventListenerSpy = jest.spyOn(document, 'addEventListener');
+
+    await asyncAction(dispatch, getState);
 
     expect(eventListenerSpy).not.toHaveBeenCalled();
     expect(actions).toContainEqual(siteLoadingUpdate(false));
--- a/src/state/actions/scigateway.actions.test.tsx
+++ b/src/state/actions/scigateway.actions.test.tsx
@@ -542,6 +542,52 @@ describe('scigateway actions', () => {
     expect(actions).toContainEqual(siteLoadingUpdate(false));
   });
 
+  it("dispatches a site loading update but doesn't wait for register route event if it's already been recieved", async () => {
+    (mockAxios.get as jest.Mock).mockImplementation(() =>
+      Promise.resolve({
+        data: {
+          'ui-strings': 'res/default.json',
+        },
+      })
+    );
+
+    const asyncAction = configureSite();
+    const actions: Action[] = [];
+    const dispatch = (action: Action): void | Promise<void> => {
+      if (typeof action === 'function') {
+        action(dispatch);
+        return Promise.resolve();
+      } else {
+        actions.push(action);
+      }
+    };
+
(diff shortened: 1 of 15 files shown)


joelvdavies:
Download and search load fine for me, but all the dataview ones give errors:
image

pluginLoadingIssue

louise-davies:
How odd - you can see I'm not messing around with anything to do with session IDs or anything - are you sure your dataview is pointing at the right servers etc.? Because I'm not seeing these errors in ISIS prod

louise-davies:
@joelvdavies I don't expect you to fix/review this - I can get Sam to do so next week, but for reference I am unable to recreate as shown below

verifying-joels-bug

src/state/actions/scigateway.actions.test.tsx
```diff
@@ -542,6 +542,52 @@ describe('scigateway actions', () => {
...
+        location: { ...createLocation('/test'), query: {} },
+        action: 'PUSH',
+      },
+    });
+
+    await asyncAction(dispatch, getState);
+
+    const eventListenerSpy = jest.spyOn(document, 'addEventListener');
+
+    expect(eventListenerSpy).not.toHaveBeenCalled();
```
sam-glendenning, line 587: Doesn't a spy need to be initialised before the tested function is called? Otherwise it would return true for not called, right? Granted, this is an async function so maybe you'd get mixed results

sam-glendenning, line 587: Doesn't a spy need to be initialised before the tested function is called? Otherwise it would return true for not called, right? Granted, this is an async function so maybe you'd get mixed results

louise-davies: We have initialised the spy - we just haven't mocked the implementation because we don't need to. I verified that this does work by adding the reverse assertion in one of the tests where we expect it to have been called. See the documentation here: https://jestjs.io/docs/jest-object#jestspyonobject-methodname

louise-davies: Oops nevermind - I see that I only spy on the method _after_ calling `asyncAction` - I'll move it to before. Good spot!

louise-davies:
@sam-glendenning since you only saw it once and it was using the dev server and not a production server, and Alejandra (who was having the issue quite regularly) says it's not happening any more I'm happy to merge this in and say it's solved - we can always look into it again later on if what you saw is consistent and affects prod.

PR #993, opened 2022-03-16
Adding Contact Us form for accessibility enquiries #990

### LINKED ISSUE #990: Investigate adding contact us form
**Description**:
One of the pieces of feedback from the review of the accessibility statement was that it would be good to have a contact us form. Dylan recommended we could use an embedded MS Form (which is what he used). Let's investigate this.

**Acceptance criteria**:
- [ ] Investigate adding contact us form (MS Form, or other solution)

sam-glendenning:
## Description
I have successfully created two separate Contact Us forms for the accessibility page using Microsoft Forms, one for ISIS and one for Diamond. However, it comes with a lot of external reliance. These are the biggest problems and we'll need to think about whether we want to fix these somehow or just work with them as they are:

- The form is an embedded Microsoft Form, owned and managed by **my** Microsoft account. If the form is ever deleted or access to it is lost for whatever reason, this would invalidate the `src` for the `<iframe>` form object, which creates an `<iframe>` with a 404 Not Found DataGateway page
- Only the form owner receives the responses from the form, meaning by default I receive the responses at sam.glendenning@stfc.ac.uk. To bypass this, I've created two flows in **Microsoft Power Automate** to forward the response to an email of my choice instead. _Currently, for testing purposes, the responses get forwarded to my personal email address_. When SciGateway goes into production, I will need to change the Power Automate flows to forward to the ISIS and DLS support emails.

The `src` value for the Contact Us form is added in SciGateway's `public/settings.json` as a field called `"contactUsAccessibilityFormUrl"`. Each form reports to separate email addresses, i.e. the ISIS form reports to the ISIS support email and the DLS form reports to the DLS support email. We'd need to make note of the exact URLs for the forms so that we don't lose them. Not sure where we'd store these. These are the links to the forms (add these into the `settings.json` file:

- ISIS: "https://forms.office.com/Pages/ResponsePage.aspx?id=HDZmP36oWEGPYZnoLbPKyO8uU1PmT5xLp0L7-EJWMzBUMUw3WjNJSzY1M09DUFo0VlNHN0w0M1M5My4u&embed=true"
- DLS: "https://forms.office.com/Pages/ResponsePage.aspx?id=HDZmP36oWEGPYZnoLbPKyO8uU1PmT5xLp0L7-EJWMzBUQzFBWDdDNzNGQzhJWDBBRk5CODZBODBNMS4u&embed=true"

I've tried my best to avert the problems I've listed above. If `"contactUsAccessibilityFormUrl"` does not exist in the settings, the form is replaced by a `mailto` email address. This email address is whatever value is set in `accessibility-page.contact-info` in `res/default.json`. This should be an ISIS email for ISIS deployment and DLS email for DLS deployment. However, this does nothing to address the problem where an invalid form URL will break the `<iframe>` entirely (any ideas for this are welcome).

If anyone could give some ideas on this functionality being improved and how to bypass some of the problems, that would be valuable.

## Screenshots
Form submission in action
Form submission compressed

The received email response
Received email

The ISIS Power Automate flow
Power Automate flow

## Testing instructions
- [x] Review code
- [x] Check Actions build
- [x] Set the `"contactUsAccessibilityFormUrl"` field in `public/settings.json` to either the ISIS or DLS URL. Load SciGateway and navigate to `/accessibility` to see the form in action. You are welcome to test that the form submission works but remember that the responses are currently being forwarded to my personal email address (can ask me to see if I received the response)
- [x] Test that the `mailto` email address for support appears in place of the form if `"contactUsAccessibilityFormUrl"` is ~~not defined~~ equal to '' (empty string) in `public/settings.json`

## Agile board tracking
Closes #990

### DIFF
--- a/src/accessibilityPage/accessibilityPage.component.test.tsx
+++ b/src/accessibilityPage/accessibilityPage.component.test.tsx
@@ -41,7 +41,7 @@ describe('Accessibility page component', () => {
 
   const theme = buildTheme(false);
 
-  it('should render correctly', () => {
+  it('should render correctly and display contact us component', () => {
     const testStore = mockStore(state);
 
     const wrapper = mount(
@@ -53,19 +53,6 @@ describe('Accessibility page component', () => {
     );
 
     expect(wrapper.find('#accessibility-page')).toBeTruthy();
-  });
-
-  it('should render contact us component correctly', () => {
-    const testStore = mockStore(state);
-
-    const wrapper = mount(
-      <Provider store={testStore}>
-        <MuiThemeProvider theme={theme}>
-          <AccessibiiltyPageWithStyles {...props} />
-        </MuiThemeProvider>
-      </Provider>
-    );
-
     expect(wrapper.find('#contact-us')).toBeTruthy();
   });
 });
--- a/src/accessibilityPage/accessibilityPage.component.test.tsx
+++ b/src/accessibilityPage/accessibilityPage.component.test.tsx
@@ -6,6 +6,12 @@ import {
 } from './accessibilityPage.component';
 import { MuiThemeProvider } from '@material-ui/core';
 import { buildTheme } from '../theming';
+import configureStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import { StateType } from '../state/state.types';
+import { authState, initialState } from '../state/reducers/scigateway.reducer';
+import { createLocation } from 'history';
+import { Provider } from 'react-redux';
 
 const dummyClasses = {
   root: 'root-class',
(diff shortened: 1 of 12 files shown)


src/accessibilityPage/accessibilityPage.component.test.tsx
```diff
@@ -16,25 +22,50 @@ const dummyClasses = {
...
+
+    const wrapper = mount(
+      <Provider store={testStore}>
+        <MuiThemeProvider theme={theme}>
+          <AccessibiiltyPageWithStyles {...props} />
+        </MuiThemeProvider>
+      </Provider>
+    );
+
+    expect(wrapper.find('#contact-us')).toBeTruthy();
```
louise-davies, line 69: Could we not just have this assertion in the previous test? Since we aren't setting up the state of the test differently

louise-davies:
Looks good and works well - my only comment is that could we add ` "contactUsAccessibilityFormUrl": "",` to the `settings.example.json` as that effectively "documents" the feature whilst having the same functionality as if it's not in the JSON at all.

src/accessibilityPage/accessibilityPage.component.test.tsx
sam-glendenning: I'll make the change

sam-glendenning:
> Ah, I didn't mean to make it mandatory! Just that it should be in the example settings file! The existed code worked fine, and I wouldn't expect SG to break if we excluded the setting...

Whoops... I'll be right back.

louise-davies (approved):

PR #1040, opened 2022-04-20
Create Actions workflow for uploading container image to Harbor

### LINKED ISSUE #1030: Create Actions workflow for uploading container image to Harbor
**Description**:
An Actions workflow needs to be created that will push the Docker container image to Harbor. The workflow from DataGateway can be used as a guide https://github.com/ral-facilities/datagateway/blob/feature/k8s-deployment/.github/workflows/docker-build.yml

**Acceptance criteria**:
- [ ] Successful Harbor upload

VKTB:
## Description
Creates a [GitHub Actions workflow](https://github.com/ral-facilities/scigateway/actions/workflows/docker-build.yml) for building and pushing the `Dockerfile` to the SCD Harbor Docker registry. It uses a Harbor robot account whose credentials are stored as secrets in this repo.

## Testing instructions
- [x] Review code
- [x] Check Actions build

## Agile board tracking
closes #1030

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -13,3 +13,9 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v2
+      - name: Log in to Harbor
+        uses: docker/login-action@v1
+        with:
+          registry: harbor.stfc.ac.uk/datagateway
+          username: ${{ secrets.HARBOR_USERNAME }}
+          password: ${{ secrets.HARBOR_TOKEN }}
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -19,3 +19,8 @@ jobs:
           registry: harbor.stfc.ac.uk/datagateway
           username: ${{ secrets.HARBOR_USERNAME }}
           password: ${{ secrets.HARBOR_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: harbor.stfc.ac.uk/datagateway/scigateway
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,38 @@
...
+# It uses a harbor robot account, who's credentials are 
+# stored as secrets in this repo.
+
+name: Build & Push Docker Image
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - deployment/push-docker-image-to-harbor-#1030
```
moonraker595, line 12: Looks good. I think pushing branch `push-docker-image-to-harbor-#1030`  is fine for testing purposes, but at some stage we should standardize all the action files to only push the `k8s-deployment` branch, or maybe anything under deployment, so: `/deployment/*`. This would mean we'd need to rename the [plugin branch](https://github.com/ral-facilities/datagateway/tree/feature/k8s-deployment) though, as it's still using the `feature` tag.

moonraker595 (approved):

moonraker595:
I'm not sure why the tests are failing? its that something we've done?

moonraker595 (commented):
Not sure why tests are failing?

moonraker595 (approved):

VKTB:
> If this is the case we may have to run the old workflow first and attach ours to the end of that one.

@moonraker595 Regarding this, I had a look at whether the `docker-build.yml` workflow can be executed only after `ci-build.yml` workflow succeeds and this only seems to be possible on default branches (see `workflow_run` event [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) for more). Considering that  `develop` is the default branch and we are using the `k8s-deployment` as our "main" K8S branch then this will not work.

Maybe the `docker-build.yml` workflow should instead be a job in the `ci-build.yml`. That way we can easily achieve this.

VKTB:
> Maybe the docker-build.yml workflow should instead be a job in the ci-build.yml? That way we can easily achieve this.

I will investigate this separately so this PR can be merge now.

PR #1044, opened 2022-04-25
Preventing app bar from opening nav drawer on failed login attempt #999

### LINKED ISSUE #999: Failed login attempt opens navigation drawer
**Description**:
When the navigation drawer is closed and the user tries to login and fails, the navigation drawer opens. This can be demonstrated on ISIS prod right now:

Failed login

Had a brief look at the code and couldn't immediately spot what was going wrong. The reducer method for an unsuccessful login sets `state.drawerOpen` to false so this must be happening elsewhere.

**Acceptance criteria**:
- [ ] Navigation drawer remains untouched after a failed login attempt

sam-glendenning:
## Description
This was happening whenever the user would try and login and fail. If autoLogin is available, the autoLogin will login, which then triggers a useEffect in mainAppBar which causes the nav drawer to open.

This same hook causes the nav drawer to open whenever the user refreshes their browser on any other page. So I've disabled this behaviour on the login page to get the correct functionality to occur.

Ultimately, this was a bit of a faff to get working but I don't think it's messed with any of the current nav drawer logic (I tested what I could and found it to mimic how it works on ISIS prod right now). 

I've added and amended some e2e tests to replicate how things currently work on ISIS prod. There was a limited amount of testing I could do with this - our current e2e setup means there is no nav drawer to test until the user logs in. So I've done my best to test the functionality works through redirects and reloads.

## Testing instructions
Try a bunch of successful and unsuccessful login attempts. Whatever state the drawer was in, it should stay that way upon clicking sign in. Also check to make sure that whenever you refresh the page on any other page besides login, the nav drawer should be open (this is existing functionality and I had accidentally touched this in the process of making this PR so basically check that nav drawer behaviour is the same as on prod in this case).

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #999


louise-davies (approved):

louise-davies:
seems to work well :)

PR #1050, opened 2022-04-29
Homepage accessible regardless of login status #1049

### LINKED ISSUE #1049: Homepage / login options for DLS
**Description**:
Currently, we have a login page as the homepage for DLS (as it was in Topcat). I think it would be better to show the homepage and provide the login page when the user picks any of the menu items (rather than only for 'My Data' as it is for ISIS).

**Acceptance criteria**:
- [ ] change homepage and login options for DLS instance

sam-glendenning:
## Description
The user's login state now has no effect on access to the homepage or display of the navigation drawer. This means that, like ISIS, DLS now allows access to the homepage at the web root and the navigation drawer shows all web app routes (but no access allowed while not logged in). This is in comparison to how things used to work on DLS, whereby all nav drawer routes including the homepage were inaccessible until the user logged in.

This PR also includes a fix to a problem where the user would not be redirected to their original request following a successful login, instead being redirected to the homepage. This took a lot of trial and error but I came to the conclusion that this was because the current method detailed below no longer works:

- User tries to navigate to `/protected/domain`
- Router redirects to /login while storing `/protected/domain` in router state
- User successfully logs in
- A router location change is triggered to `/logout`, clearing the router state in the process (still not sure why this location change happens)
- Router state is gone at the crucial stage, therefore default redirect to web root occurs

The state is cleared between page refreshes/redirects because this is what `react-router-dom` does. But I'm unsure as to why a router `REPLACE` to `/logout` happens.

To get around this, I made a new method whereby the referral URL is stored in `localStorage` instead. It is then fetched, changed and deleted as required.

## Testing instructions
Run this with at least one other plugin loaded so that the navigation drawer can be populated. Set the `"authUrl"` in settings.json to be `"https://datagateway-preprod.diamond.ac.uk/api"` to mimic DLS DG environment (no auto login).
When navigating to the web root, you should be directed to the homepage instead of `/login`. Also, the navigation bar should be accessible but you should be redirected to `/login` upon trying to navigate to any of the routes.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1049

### DIFF
--- a/src/state/actions/scigateway.actions.tsx
+++ b/src/state/actions/scigateway.actions.tsx
@@ -490,15 +490,10 @@ export const verifyUsernameAndPassword = (
         dispatch(authorised());
 
         // redirect the user to the original page they were trying to get to
-        // the referrer is added by the redirect in routing.component.tsx
-        const previousRouteState = getState().router.location.state;
-        dispatch(
-          push(
-            previousRouteState && previousRouteState.referrer
-              ? previousRouteState.referrer
-              : '/'
-          )
-        );
+        // the referrer is set in localStorage in authorisedRoute.component.tsx
+        const referrer = localStorage.getItem('referrer');
+        localStorage.removeItem('referrer');
+        dispatch(push(referrer ?? '/'));
       })
       .catch(() => {
         // probably want to do something smarter with
--- a/src/state/actions/scigateway.actions.tsx
+++ b/src/state/actions/scigateway.actions.tsx
@@ -485,14 +485,14 @@ export const verifyUsernameAndPassword = (
     await authProvider
       .logIn(username, password)
       .then(() => {
+        const referrer = getState().router.location.state?.referrer;
+
         if (newMnemonic)
           dispatch(loadAuthProvider(`icat.${newMnemonic}`, `${authUrl}`));
         dispatch(authorised());
 
         // redirect the user to the original page they were trying to get to
-        // the referrer is set in localStorage in authorisedRoute.component.tsx
-        const referrer = localStorage.getItem('referrer');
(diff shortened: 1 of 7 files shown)


src/state/actions/scigateway.actions.tsx
```diff
@@ -490,15 +490,10 @@ export const verifyUsernameAndPassword = (
         dispatch(authorised());
 
         // redirect the user to the original page they were trying to get to
-        // the referrer is added by the redirect in routing.component.tsx
-        const previousRouteState = getState().router.location.state;
```
louise-davies, line 494: I think if you moved the reading of `previousRouteState` before dispatching the `authorised()` action then the router referrer would be accurate. The redirect to `/logout` happens because if a user tries to access the login page when logged in they should instead see the logout page. If you grab what's in the referrer before the user is logged in, then this should be before the redirect to `/logout`.

Though tbh, I don't mind doing it via localStorage, just thought you'd want to know what was causing the problem :)

sam-glendenning: Oh I think I get it. I'll give it another go

louise-davies (commented):
This works well - on page load you see the homepage, and upon clicking any of the links on the homepage or in the sidebar you are redirected to the login page and once you login you are taken back to the link you clicked on and the sidebar opens as a default.

I'll wait to see if you manage to get the old redirect working though

sam-glendenning:
@louise-davies fixed with your solution! You were right, I simply needed to grab the referrer from state a little earlier. Thanks for the help

louise-davies (approved):
Works well with the new fix - glad I could help out and that it worked :)

PR #1064, opened 2022-05-17
Improving and fixing accessibility statement

sam-glendenning:
## Description
While reviewing the React 17 upgrade, I noticed a couple of things that could be improved about the accessibility statement, including one statement that was incorrect and needs to be fixed (line 177). I also need some clarification on line 175, particularly what it means and whether it needs to be changed to "...elements that _do_ contain focusable elements".

## Testing instructions
Double check the language changes I've made and confirm you're happy with them.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
No issue as these are minor changes made on the fly

### DIFF
--- a/public/res/default.json
+++ b/public/res/default.json
@@ -145,9 +145,9 @@
     "non-accessibile-parts-of-datagateway-list": [
       "the columns on table view cannot be resized using the keyboard",
       "pages sometimes have unreadable ARIA labels",
-      "table view has aria hidden elements that contain focusable elements and elements in the table with an ARIA role that required a parent are not contained by them ",
+      "table view has aria-hidden elements that contain focusable elements and elements in the table with an ARIA role that require a parent are not contained by them ",
       "nested interactive controls are not announced by screen reader in the sorting in card view ",
-      "advanced filters, filters automatically without requiring user input"
+      "advanced filters activate automatically without requiring user input"
     ],
     "contact-us-about-accessibility": "Contact Us About Accessibility",
     "contact-us-about-accessibility-text": "We are always looking to improve the accessibility of this website. If you find any problems not listed on this page or think we are not meeting accessibility requirements, please let us know. You can send us a message to request a more accessible format for content or documents on the site. ",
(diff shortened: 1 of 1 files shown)


public/res/default.json
```diff
@@ -165,16 +165,16 @@
...
-      "The advanced filters, filters automatically without requiring user input. This fails WCAG 2.1 success criterion 3.2.2 On Input"
+      "The advanced filters activate automatically without requiring user input. This fails WCAG 2.1 success criterion 3.2.2 On Input"
     ],
     "non-compliance-with-the-accessibility-regulations-text": "We will make sure that all the non-compliances above are addressed before the next release, likely to be in September 2022.",
     "disproportionate-burden": {
       "title": "Disproportionate burden",
       "table-view": "Table view",
       "table-view-list": [
-        "There is no way to have ARIA hidden elements that do not contain focusable elements in the text field of the filters",
+        "There is no way to have aria-hidden elements that do not contain focusable elements in the text field of the filters",
```
sam-glendenning, line 175: Line 175 that I still need confirmation on

louise-davies:
@sam-glendenning re: line 175 - I think maybe the better wording is (as the context of that section is things we aren't/can't fix) `It is not possible for us to remove the aria-hidden elements that contain focusable elements in the text field of the filters` - i.e. the offence is that there are `aria-hidden` elements with focusable elements in them - so we are trying to say it's a disproportionate burden to try and fix that. I'll need to double check the context of this after everything comes back online after this weekend

sam-glendenning:
Ah, I think I see now. So the problem is that the text filter has aria hidden content that shouldn't be focusable but is. And we can't do anything about it (or at least haven't found a way). Going by your last sentence, I'll wait on this for the moment before taking any action

louise-davies:
@sam-glendenning I've just checked, and I think the wording is `It is not possible for us to remove the aria-hidden attribute from a  focusable element in the text field of a text filter`. Otherwise, I think the rest of your suggestions can go in as is

sam-glendenning:
@louise-davies sorry I'm only just getting back to this, does this look good with that most recent change?

louise-davies (changes requested):
The new content is fine - but it looks like in the merge conflict resolution you overwrote some changes (e.g. the cookie page change) - can those changes be reverted so they match what's on `develop`?

sam-glendenning:
Thanks for spotting that, looks like I managed to totally screw up that merge from develop! Should be better now?

public/res/default.json
```diff
@@ -75,7 +75,7 @@
...
...dow will submit your download request to be processed. You should see a <strong>Cart successfully submitted</strong> notification popup on the right corner. The request will be added inside the downloads window.<h2 id='download'> Download</h2>Clicking on the above will open up the downloads window:<br><img src='/res/images/downloads-button.png'><br>The Downloads window is a grid that list all your download requests.<br><img src='/res/images/downloads.png'><br>The <strong>Status</strong> column gives you the current status of your download. Possible values are <strong>Restoring from Tape</strong> and <strong>Available</strong>. The <strong>Download</strong> button is disabled when the status is restoring. Once available, you can click the download button to download the file via the browser if <strong>Https</strong> transport type was used for that request.<br><br>Please note that download pausing and resuming is not supported. Multi-part download managers are also not supported.<br><br>For downloads using Globus as the transport type, when the status is complete, you can login to <a href='https://www.globus.org/SignIn'>Globus Online</a> and initiate file transfer to your endpoint."
```
sam-glendenning, line 78: One extra thing I did do is change TopCAT to DataGateway (it is TopCAT on develop for some reason). If there's a reason this is still TopCAT on develop I'll change it back but just wanted to point this out in case you missed it

louise-davies (approved):
Looks good, and the merge conflict problems have been fixed

PR #1072, opened 2022-05-24
Fix token refresh error

louise-davies:
## Description
Goes with ral-facilities/datagateway#1262

Allow plugins to send a notification payload alongside their token invalidation request event, which we then only display after the token refresh has failed.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Get me to lower the ICAT session length (or mock that yourself) and test with PR ral-facilities/datagateway#1262

## Agile board tracking
Part of the fix for ral-facilities/datagateway#1257

### DIFF
--- a/public/res/default.json
+++ b/public/res/default.json
@@ -67,35 +67,6 @@
       "button": "Read more"
     }
   },
-  "home_page": {
-    "title_line1": "<0>Data discovery</0> and <2>access</2>",
-    "title_line2": "for <1>large-scale</1> science facilities",
-    "browse": {
-      "title": "Browse, explore and visualise experimental data",
-      "description1": "Large scale facilities, such as synchrotrons, neutron and muon sources, lasers and accelerators, generate vast amounts of data that need to be managed in an efficient way, supporting data ingestion for long-term storage and archival, as well as data analysis and data publication workflows.",
-      "description2": "<0>DataGateway</0> focuses on providing data discovery and data access functionality to the data.",
-      "link": "/browse/investigation",
-      "button": "Browse data"
-    },
-    "search": {
-      "title": "Search",
-      "description": "Search for the experimental data according to different criteria.",
-      "link": "/search/data",
-      "button": "Search data"
-    },
-    "download": {
-      "title": "Download",
-      "description": "Retrieve the experimental data using a variety of download methods.",
-      "link": "/download",
-      "button": "Download data"
-    },
-    "facility": {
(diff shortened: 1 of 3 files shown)


src/state/middleware/scigateway.middleware.tsx
```diff
@@ -270,9 +286,25 @@ export const autoLoginMiddleware: Middleware<
...
+        .catch(() => {
+          log.error('Auto Login via middleware failed');
+          // we can't recover from here - tell user they'll need to refresh the page or login again
+          document.dispatchEvent(
+            new CustomEvent(microFrontendMessageId, {
+              detail: {
+                type: NotificationType,
+                payload: {
+                  severity: 'error',
+                  message: getString(res, 'auto-login-error-msg'),
```
sam-glendenning, line 302: Looks good from this side but I think you forgot to define the `auto-login-error-msg`. I was going to make a commit to quickly add it myself but wanted your thoughts on what the contents of this message should actually be. This would be a fairly unique situation where the auto login has failed for some reason. Is it the best idea to ask the user to "reload the page" and try again or should we advise them to do something else entirely like try logging in for real. Because if the auto login fails, there could be a bigger problem in the backend as a whole, I'm not sure.

sam-glendenning (changes requested):

src/state/middleware/scigateway.middleware.tsx
louise-davies: Good catch! I have actually wrote the message, just must have missed committing it. I'll add it myself soon so you don't need to worry about the wording. I did write a message along the lines of `please refresh the page and if the error persists contact support` or something

public/res/default.json
```diff
@@ -66,6 +67,35 @@
...
+      "link": "/download",
+      "button": "Download data"
+    },
+    "facility": {
+      "title": "ISIS Neutron and Muon Source",
+      "description": "World-leading centre for research giving unique insights into the properties of materials on the atomic scale.",
+      "link": "https://www.isis.stfc.ac.uk/Pages/About.aspx",
+      "button": "Read more"
+    }
+  },
```
sam-glendenning, line 98: Not sure how this happened... did you accidentally copy this over from somewhere?

louise-davies: Oops, I didn't realise I didn't revert my default.json before modifying it - I'll fix it now...

sam-glendenning (approved):

PR #1076, opened 2022-05-31
Adding arrow on left of headers in help page #1075

### LINKED ISSUE #1075: Help page - add arrow to go to the top of the page
**Description**:
It would be useful to have a small arrow next to each heading, so that when clicked, it takes users to the top of the page (rather than having to scroll all the way up).

**Acceptance criteria**:
- [ ] Include way to go to the top of the page next to each header, possibly adding a small arrow next to the title

sam-glendenning:
## Description
As the help page is by far the longest page, I've added an arrow next to each of the main headings in the help page which, when clicked, takes the user to the top of the page again so they don't have to scroll all the way back up. Due to the help page being constructed from pure HTML, I had to use a Unicode-style icon for the links. This also required some additional logic in the table of contents construction so the icon wouldn't appear in the header names.

I couldn't decide which side to have the arrow on so _kept it on the left for now_ but this can be easily changed. Here's what they look like for reference:

Left:
left

Right:
right

## Testing instructions
Go to /help and try it out!

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1075


louise-davies:
@sam-glendenning Could you take a look at the new code I wrote?

sam-glendenning:
Actually never mind, I have admin privileges, I'm happy with it, you're happy with it, we don't need this back and forth for a tiny PR

PR #1077, opened 2022-05-31
Implement best practices in Docker image

VKTB:
## Description
This PR implements best practices in the Docker image.

In summary, the changes do the following:
- Removes the lines that sets the `PATH` because the image builds and the container works fine without it.
- Build cache is leveraged in the build stage to reduce the build time when changes are made to the source code over and over again. This is accomplished by copying the `package.json`, `tsconfig.json` and `yarn.lock` files and installing the dependencies first, and then copying the source code and building the app. Because each `COPY` and `RUN` command is a new layer, this ensures that the layers for copying and installing the dependencies are cached when changes are made to the source code.
- Configures the image so that containers run using the non-root `www-data` user account that the `httpd` image defines.
- Uses `setcap` to allow non-root user accounts to bind to privileged ports (port 80 in this case) as these ports are only permitted to `root` by default.
- Changes the access rights for the Apache's `logs` directory from `root` to `www-data` user account.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Image builds successfully
- [ ] Container starts successfully

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+/.github
+/.husky
+/cypress
+/demos
+.eslintrc.js
+.gitignore
+.prettierrc
+codecov.yml
+CODEOWNERS
+cypress.json
+Jenkinsfile
+LICENSE.md
+README.md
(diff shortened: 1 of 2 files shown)


moonraker595 (approved):
Looks good 👍

PR #1084, opened 2022-06-09
#1081 - add autoLogin setting to control whether we attempt autoLogin

### LINKED ISSUE #1081: Error message appears to do with autologin when there's no anon authenticator
**Description**:
I added an error message related to an autoLogin feature, but this triggers on DLS which doesn't have autoLogin capability due to no ICAT anon authenticator. The error should only occur on autoLogin compatible instances.

**Acceptance criteria**:
- [ ] Auto login failure message only appears on auto login enabled instances of SciGateway

louise-davies:
## Description
See title.

Also removed the `authUrl` argument from `verifyUsernameAndPassword` because it doesn't change so we can just reuse the old value

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] No autologin error appears when using DLS authenticator and `autoLogin: false`

## Agile board tracking
Closes #1081

### DIFF
--- a/cypress/integration/login.spec.ts
+++ b/cypress/integration/login.spec.ts
@@ -303,6 +303,7 @@ describe('Login', () => {
         'ui-strings': 'res/default.json',
         'auth-provider': 'icat',
         authUrl: 'http://localhost:8000',
+        autoLogin: true,
         'help-tour-steps': [],
       });
       cy.intercept('/authenticators', [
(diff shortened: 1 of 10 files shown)


sam-glendenning (approved):
Just made a small commit to clarify some meaning in tests - apart from that I'm happy (assuming the tests pass)

PR #1087, opened 2022-06-10
Customisable primary colour

### LINKED ISSUE #1085: Allow main app bar colour to be configurable
**Description**:
See title - this is required for DLS.

**Acceptance criteria**:
- [ ] Main app bar colour can be configured

louise-davies:
## Description
Add the ability to specify a custom primary colour that is used on the app bar etc.

Note, I didn't override the dark mode high contrast mode colour, as the use case for this is DLS which they want a darker blue than we currently use `#072B5D` but if in the future we'd expect more colours then we'd need to separate out so we can customise the high contrast colours separately.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Add custom colour to your settings and see that the main app bar & footer change colour

## Agile board tracking
Closes #1085

### DIFF
--- a/public/settings.example.json
+++ b/public/settings.example.json
@@ -6,6 +6,7 @@
     "dark": "",
     "altTxt": ""
   },
+  "primaryColour": "",
   "adminPageDefaultTab": "maintenance",
   "features": {
     "showHelpPageButton": true,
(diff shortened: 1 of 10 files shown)


sam-glendenning (approved):
Works well! Customises the SciGateway side of things, as well as the breadcrumbs in DG. I noticed that specifying a colour name, such as 'red', wasn't accepted but that's okay because hex colours are more accepted and customisable anyway

louise-davies:
@sam-glendenning I think it needs a hex colour as it modifies the MUI theme colour rather than being a CSS property directly. If someone did want to use a HTML colour they could just convert to hex

PR #1090, opened 2022-06-13
#1086 - enable custom logos to be larger than 24px

### LINKED ISSUE #1086: DLS logo needs to be bigger in mainAppBar
**Description**:
I think the logo height of 24 is too small - either we set a max height of perhaps 32 (which is similar the height of the buttons iirc?) (switch to using max height so then you can choose to have your logo smaller rather than it stretching to fit). or do something crazy with max width?

**Acceptance criteria**:
- [ ] DLS logo is larger

louise-davies:
## Description
Instead, we just set a sensible max height based off of the app bar height. Default logos I set to be 24px height instead so we don't have to modify the SVGs to be smaller.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #1086

### DIFF
--- a/src/mainAppBar/mainAppBar.component.tsx
+++ b/src/mainAppBar/mainAppBar.component.tsx
@@ -84,7 +84,9 @@ const styles = (theme: Theme): StyleRules =>
       padding: 4,
       margin: theme.spacing(1),
       '& img': {
-        height: 24,
+        maxHeight: `calc(${
+          (theme as UKRITheme).mainAppBarHeight
+        } - 2 * 4px - 2 * ${theme.spacing(1)}px)`,
       },
     },
     button: {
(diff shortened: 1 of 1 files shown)


src/mainAppBar/mainAppBar.component.tsx
```diff
@@ -84,7 +84,9 @@ const styles = (theme: Theme): StyleRules =>
       padding: 4,
       margin: theme.spacing(1),
       '& img': {
-        height: 24,
+        maxHeight: `calc(${
+          (theme as UKRITheme).mainAppBarHeight
+        } - 2 * 4px - 2 * ${theme.spacing(1)}px)`,
```
sam-glendenning, line 89: How did you calculate this?

louise-davies: That shows I should add a comment! It's just minusing the margin and padding from the overall height, so the logo can take up all the remaining space. everything is times 2 because things have bottom and top padding & margin

sam-glendenning: Ah I'm with you now

sam-glendenning (approved):

PR #1121, opened 2022-07-07
Limit height of page content so that app bar and nav bar stays visible when scrolling

### LINKED ISSUE #996: Scrolling for Help and Accessibility statement pages
**Description**:
Scrolling in browse/search works so that the top of the page, menu and footer are always visible.

We should have the Help and Accessibility statement have the same behaviour.

**Acceptance criteria**:
- [ ] Help and Accessibility statement support scrolling by keeping top of the page, menu and footer.

kennethnym:
## Description
This PR limits the height of `PageContainer` so that the app bar and the footer is still visible when the page is scrolling.

## Demo

This clip shows the new scroll behavior:

https://user-images.githubusercontent.com/19359984/177801660-cd892469-c655-4210-b586-10f3a54a815e.mp4

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #996 

Closes #996


src/routing/routing.component.test.tsx
```diff
@@ -9,6 +9,8 @@ import { MemoryRouter } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import TestAuthProvider from '../authentication/testAuthProvider';
 import * as singleSpa from 'single-spa';
+import { ThemeProvider } from '@emotion/react';
```
sam-glendenning, line 12: I'm not an expert on the difference it makes but for consistency with other unit tests, you may want to import the `ThemeProvider` from `@mui/material` instead. This may even lead to a smaller snapshot change than the currently huge one that is present!

sam-glendenning (changes requested):
UI-wise it looks good but as Louise said we still have an extra vertical scroll bar present in table view. This mostly affects Dataview from what I can see but I had a look elsewhere and saw maybe a couple of other areas that could maybe be improved too.

For example, on the admin download table, we have an extra scroll bar that we probably don't need and this is because the table element is surrounded by what looks to be a Paper component. I removed it and it got rid of the problem but also caused the whole section to be moved too far up and left, like so:

image

Have a play around with it and see if you can get it lining up nicely, I think it just needs some padding somewhere.

Similarly, I noticed the same problem on the downloadTab component in DataGateway Download which I _think_ might be [caused by this](https://github.com/ral-facilities/datagateway/blob/73cd741772675501ad1b0bfde1c6d7626270f50d/packages/datagateway-download/src/downloadTab/downloadTab.component.tsx#L73)? So if you fancy having a go at that one too, that would be cool, would just require a PR on DataGateway repo.

src/routing/routing.component.test.tsx
kennethnym: Sorry, I didn't even realize, my IDE must have imported it from the wrong package!

kennethnym:
Thank you for the review! I'll look into the scroll bar issue.

src/routing/routing.component.test.tsx
```diff
@@ -117,7 +120,9 @@ describe('Routing component', () => {
         <MemoryRouter
           initialEntries={[{ key: 'testKey', pathname: '/test_link' }]}
         >
-          <Routing />
+          <ThemeProvider theme={theme}>
+            <Routing />
+          </ThemeProvider>
```
sam-glendenning, line 125: Sorry, one last change! Adding the `ThemeProvider` to the mounted wrapper has resulted in some massive additions to the snapshots that we just don't need. If you take it out from all the mounted wrappers in this file, you should get some more manageable snapshots. Once it's done, compare the new snapshots to what is currently on develop branch and as long as they look fairly similar in size, they're probably good to go!

kennethnym: No worries! I didn't know it wasn't needed. I thought the `Routing` component would need it but turns out it doesn't. I don't think it should be removed though as we will have to rewrite it in `@testing-library` in the future which will actually require `ThemeProvider` since it renders the entire component out, and all the styling obviously requires the `theme` object to be in context.

sam-glendenning:
Also I think something might have gone wrong with the table in dataview... the bottom has been cut off

image

src/routing/routing.component.test.tsx
sam-glendenning: Personally, if we remove it, we can more easily test the snapshot is correct during the migration to testing-library, plus just leaving the import by itself will make the linter complain. But that's just my opinion

kennethnym:
I think it is due to a slight miscalculation of height in `datagateway-dataview`. I played around with the number and I found that replacing `calc(100vh - 180px - 36px - 1px - ${linearProgressHeight} - ${breadcrumbHeight})` with `calc(100vh - 152px -36px - 1px - ${linearProgressHeight} - ${breadcrumbHeight})` makes the container fit perfectly, although I don't really know why it is specifically 152px...

sam-glendenning:
I thought it could be something like that. Try the solution on a bunch of pages before submitting it, just to make sure it doesn't accidentally break something else!

kennethnym:
The calculation is only for `datagateway-dataview`'s table (specifically [this line](https://github.com/ral-facilities/datagateway/blob/7324fc0f797a0b79adf9aa3935d0be5da9dec077/packages/datagateway-dataview/src/page/pageContainer.component.tsx#L389), so it shouldn't affect other pages, but I viewed other pages normally and in the responsive tool and couldn't produce any regression.

sam-glendenning (approved):
Tested this with https://github.com/ral-facilities/datagateway/pull/1378 and it works well!

We do now have a snapshot that's bigger than it's worth but as long as we put a pin in it to remember to reduce it when we refactor tests, it can stay for now

kennethnym:
It should become smaller as we slowly migrate to `@testing-library` because instead of snapshotting the theme object, it snapshots the actual HTML output as if it were in a browser.

PR #1126, opened 2022-07-11
Update yarn.lock

louise-davies:
## Description
Due to the MUiv5 branch, some security fixes to yarn.lock were lost, and since dependabot doens't work with yarn v2 I thought I'd just manually fix these.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3848,11 +3848,11 @@ __metadata:
   linkType: hard
 
 "async@npm:^2.6.2":
-  version: 2.6.3
-  resolution: "async@npm:2.6.3"
+  version: 2.6.4
+  resolution: "async@npm:2.6.4"
   dependencies:
     lodash: ^4.17.14
-  checksum: 5e5561ff8fca807e88738533d620488ac03a5c43fce6c937451f7e35f943d33ad06c24af3f681a48cca3d2b0002b3118faff0a128dc89438a9bf0226f712c499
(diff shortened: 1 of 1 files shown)


sam-glendenning:
Could you give an idea of what you've done, how I can verify this, etc.?

louise-davies:
I ran:

`npm_config_yes=true npx yarn-audit-fix` (from https://www.npmjs.com/package/yarn-audit-fix)

and then ran

`yarn dedupe` (from https://yarnpkg.com/cli/dedupe/)

You can run `yarn audit -R -A` (-R: recursively check dependencies, -A: check all package.json files) to see that there should be only 1 problematic dependency left, which is nth-check which is unfixable as it's required by `react-scripts`

louise-davies:
However, as I said we had already fixed these dependencies in the Diamond release - this is just added them back in as they were borked by the MUIv5 branch (as they only updated the lockfile) and I discovered what `yarn dedupe` is which is a completely safe operation.

sam-glendenning:
@louise-davies ah I think I see, I presume you're referring to this, which I had no idea about
https://github.com/facebook/create-react-app/issues/11174#issue-935928547

So just one final comment, based on that link - is it worth moving `react-scripts` into `devDependencies`?

louise-davies:
@sam-glendenning unfortunately that would break the workflow that Alex & Viktor want for the docker stuff - as `react-scripts` is needed to build SG it needs to be a normal dependency

PR #1212, opened 2022-10-31
Removing @types/node

sam-glendenning:
## Description
Unneeded dependency? 

## Testing instructions
Dependency is gone. Related to https://github.com/ral-facilities/scigateway/pull/1210

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Hotfix

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,6 @@
     "@testing-library/user-event": "14.4.2",
     "@types/history": "4.7.3",
     "@types/jsonwebtoken": "8.5.0",
-    "@types/node": "17.0.17",
     "@types/react": "17.0.38",
     "@types/react-redux": "7.1.20",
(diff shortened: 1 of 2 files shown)


louise-davies:
@sam-glendenning I'm not certain why the dependency is here, but it's one added by default by CRA and it's still recommended: https://create-react-app.dev/docs/adding-typescript/. I think it would theoretically get used in the little servers e.g. the e2e server, the example auth server as well. I'd say keep it in as it's only a dev dependency and it's still recommended by CRA

sam-glendenning:
Thanks for the clarification. Funnily enough, I was just updating a package or two and noticed this one suddenly appearing a lot more in yarn.lock. So clearly it is being used in some fashion. I'll close this PR

PR #1228, opened 2022-11-23
 #122 - have CDN fallbacks for react & react-dom

### LINKED ISSUE #122: Reconsider CDN usage?
**Description**:
We are currently using `unpkg` to include React via CDN. However, this has proved to be relatively unreliable as there has been two instances of React unable to be reached via our apps. I think we should either switch to a more reliable CDN (`jsDelivr` or `cdnjs` perhaps?) or host React ourselves (production only, development should probably still use CDN).

**Acceptance criteria**:
- [ ] Decision on which CDN to use (or usage of none at all) is made and documented as an ADR

louise-davies:
## Description
Add some commands to the postinstall script to put copies of react & react-dom in the public folder, and then in `index.html` refer to those files if the CDN files fail to load.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check that public folder contains react & react-dom production files after install (make sure that your `yarn install` actually installs new stuff though as otherwise the postinstall won't run - I don't think this is a problem for CI (and thus the CI release process) though as it will always be a fresh install)
- [ ] Block the requests from the CDN and test that it loads react&react-dom from localhost instead

## Agile board tracking
Closes #122


sam-glendenning (changes requested):
Seems to work pretty well! On Linux... the `cp` command is not valid on Windows. We might need a different solution here so that `yarn install` doesn't fail all the time on Windows. We could potentially use something like [this](https://www.npmjs.com/package/copyfiles) or [this](https://www.npmjs.com/package/copy-files-from-to)?

louise-davies:
@sam-glendenning dope! Forgot that `cp` is linux only! I'll figure something out - to avoid an extra dependency I might just make a simple node.js script which uses `fs` to do the copying which is cross-platform

sam-glendenning (approved):
Yep that seems to have done it. Nice feature!

PR #1271, opened 2023-03-10
Feature/use ga4 #1219

### LINKED ISSUE #1219: Use GA4 instead of UA in google analytics
**Description**:
Universal Analytics will stop working on July 1st 2023, so we'll need to migrate to use GA4 before then. The current package we use - `react-ga` has not been updated for GA4. Someone created `react-ga4` which is meant to be a drop in replacement for `react-ga`, or we could look into alternative solutions (e.g. theoretically we don't really need a "react" library for this - perhaps it's just better to use plain JS so we're not in this situation again where GA updated but our package is dead).

**Acceptance criteria**:
- [ ] Use GA4 instead of UA for GA tracking

jounaidr:
## Description
Changes to replace react-ga package with default google analytics 4 tag.  Login events have been removed, and page view sections have been migrated and seem to be working fine. The `gtag` is initialised through DOM manipulation within [cookieConsent.component.tsx](https://github.com/ral-facilities/scigateway/pull/1271/files#diff-52c7f13d06303b5aadad5b992d16a611074d6e957fea319cc2cb6e9e77e39db8), and since page views are now sent automatically reference and use of analytics within the middleware has been removed.

## Testing instructions
Can be tested by using a GA4 id tag in the settings, and once run as localhost page views and other related data should be seen within the google analytics dashboard.

## Agile board tracking
Closes #1219


src/cookieConsent/cookieConsent.component.tsx
```diff
@@ -52,20 +51,22 @@ export const CookieConsent = (
...
-        anonymizeIp: true,
-        page,
-      });
-      // need to send initial pageview
-      ReactGA.pageview(page);
+      //Global Site Tag (gtag.js) - Google Analytics
+      const urlScript = document.createElement('script');
+      urlScript.async = true;
+      urlScript.src =
+        'https://www.googletagmanager.com/gtag/js?id=G-BMV4M8LC8J';
```
louise-davies, line 58: The `id` query param here should be grabbed from `props.analytics.id` right?

louise-davies (changes requested):
See comment

src/cookieConsent/cookieConsent.component.tsx
jounaidr: Thanks, yeh it should! changes made.

louise-davies (approved):
I've tested this and I can see it tracking by page views so this is good to go :) I'll add a card to the board to remind me that the GA properties will need migrating when this is next deployed

PR #1276, opened 2023-06-01
Containerize application and configure GitHub Actions to build and push Docker image to Harbor

VKTB:
## Description
This PR adds a Dockerfile which can be used to build a Docker image and create a container using it. It is configured to create a production image and run an Apache httpd server on port `80` when a container is started.

To add an extra layer of security, the image has been set up so that containers run as non-root.

Environment variables have also been defined in the Dockerfile to allow for values to be passed at runtime to future running containers. These values are then used by the `docker-entrypoint.sh` script, which acts as an entry point script that is executed when the container starts, to update the config values in the `settings.json` file. The environment variables are:
- `AUTH_PROVIDER` (Default value: `icat`)
- `AUTH_URL` (Default value: `/api`)

A separate `settings.json` file is introduced which is configured to use the 3 different DataGateway plugins and the Dockerfile is set to use this instead of creating one from the `settings.example.json` file. The reason for this is because the example file uses the demo plugin and it would be too complicated to allow for the list of plugins to be configured through ENV variables.

Finally, this PR adds a `docker` job to the `ci-build` GitHub Actions workflow to build a Docker image using the Dockerfile introduced in this PR and push the same to Harbor. The job only triggers if all the other jobs succeed and does different things depending on the context. It builds the Docker image in all cases and also pushes the image to Harbor only if something is pushed to the `main` or `develop` branch.

To build an image, run the command below in the `scigateway` repo directory
```
docker build -t scigateway_image .
```
To start a container on port `80` from the image that you just built, run:
```
docker run -p 80:80 --name scigateway_container scigateway_image 
```
If you want to pass values for the environment variables then instead run:
```
docker run -p 80:80 --name scigateway_container --env AUTH_PROVIDER=jwt --env AUTH_URL=/auth-api scigateway_image
```

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] GitHub Actions should build and push an image to Harbor when this PR is merged


.dockerignore
VKTB, line 1: Are there any other files or directories that are not needed when creating a `build`? If so we can add them to this list to exclude them from being copied.

.github/workflows/ci-build.yml
```diff
@@ -34,3 +33,35 @@ jobs:
...
+          username: ${{ secrets.HARBOR_USERNAME }}
+          password: ${{ secrets.HARBOR_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: harbor.stfc.ac.uk/datagateway/scigateway
+
+      - name: ${{ github.event_name == 'push' && github.ref == 'refs/heads/develop' && 'Build and push Docker image to Harbor' || 'Build Docker image' }}
```
VKTB, line 61: From what I remember, I think we also want to push an image when stuff gets merged into `main` as well?

.dockerignore
ajkyffin: It doesn't matter if unnecessary files get copied., as long as Dockerfile is in there so you can change it without invalidating the whole build.

.github/workflows/ci-build.yml
ajkyffin: Yes, otherwise we can't edit the Dockerfile without merging in all the changes from develop.

ajkyffin (approved):
Squash-merge this because there are so many commits.

ajkyffin (dismissed):

louise-davies (approved):
LGTM - I deployed this on a cloud machine and it worked fine other than the issue I reported on Gitlab (which I suppose could be fixed on this PR or could be fixed with another PR - up to you)

VKTB:
> LGTM - I deployed this on a cloud machine and it worked fine other than the issue I reported on Gitlab (which I suppose could be fixed on this PR or could be fixed with another PR - up to you)

@louise-davies I fixed the caching and page refresh issues as part of the last 3 commits.

louise-davies (commented):
The caching doesn't work when loaded via the docker compose file (where everything is routed via a nginx server). Might only need modifications in that file which is in a separate repo, hence not necessicarily requesting changes on this PR

VKTB:
> The caching doesn't work when loaded via the docker compose file (where everything is routed via a nginx server). Might only need modifications in that file which is in a separate repo, hence not necessicarily requesting changes on this PR

As discussed, caching work for me locally. The reason I think it didn't work when you tested was because the Harbor image didn't get updated meaning that it didn't have the cache fix when you pulled the image. I have modified the Docker GitHub Actions job to push the image during PRs.

ajkyffin (approved):

.dockerignore
louise-davies, line 1: Probably want to ignore `**/node_modules` and maybe also `/build` - ensures the Docker image isn't potentially polluted by the local dev environment

louise-davies: Also `/coverage` can reduce the size of images on developer machines - shouldn't affect CI build though

.dockerignore
```diff
@@ -0,0 +1,13 @@
...
+/demos
+.gitignore
+.prettierrc
+codecov.yml
+CODEOWNERS
+cypress.json
+Jenkinsfile
+LICENSE.md
+README.md
+Dockerfile
```
VKTB, line 13:
```suggestion
/.github
/.husky
/build
/coverage
/cypress
/demos
**/node_modules
.gitignore
.prettierrc
codecov.yml
CODEOWNERS
cypress.json
Jenkinsfile
LICENSE.md
README.md
Dockerfile
```

.dockerignore
VKTB: Thanks! Addressed in https://github.com/ral-facilities/scigateway/pull/1276/commits/154b570472002abfc478f78b868748b68ac42e18.

louise-davies (approved):
LGTM

PR #1314, opened 2023-10-04
Add severity selection to admin scheduledMaintenance #1053

kaperoo:
## Added severity selection in admin scheduledMaintenance interface
It is now possible to choose a severity for the scheduled maintenance pop-up in the admin interface. 
image

Added a new 'information' severity. 
image

## Testing instructions
Added unit tests to include cases for the 'information' in invalid severity. Updated tests to select severity from the selection list.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1053

### DIFF
--- a/src/state/actions/scigateway.actions.tsx
+++ b/src/state/actions/scigateway.actions.tsx
@@ -439,6 +439,9 @@ export const configureSite = (): ThunkResult<Promise<void>> => {
       provider
         .fetchScheduledMaintenanceState()
         .then((scheduledMaintenanceState) => {
+          if (scheduledMaintenanceState['severity'] === undefined) {
+            scheduledMaintenanceState['severity'] = 'warning';
+          }
           dispatch(loadScheduledMaintenanceState(scheduledMaintenanceState));
 
           // Checking the state in the GET response because it does not get
--- a/src/state/actions/scigateway.actions.tsx
+++ b/src/state/actions/scigateway.actions.tsx
@@ -446,7 +446,7 @@ export const configureSite = (): ThunkResult<Promise<void>> => {
           if (scheduledMaintenanceState['show']) {
             displayMaintenanceBanner(
               scheduledMaintenanceState['message'],
-              'warning'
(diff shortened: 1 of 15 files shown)


src/state/scigateway.types.tsx
```diff
@@ -188,6 +188,7 @@ export interface ScheduledMaintenanceStatePayLoad {
 export interface ScheduledMaintenanceState {
   show: boolean;
   message: string;
+  severity: 'success' | 'warning' | 'error' | 'information';
```
louise-davies, line 191: For backwards compatibility - we should probably allow this to be undefined as well, and handle that logic to assume it means warning as otherwise this could cause problems when upgrading. For example, deleting the `severity` from the SG preprod backend JSON file means a toast won't display - despite `show` being true. All the facility preprod/prod machines don't have `severity` in the JSON.

On the other hand, if it makes the code more complicated unecessarily, we can change https://github.com/ral-facilities/scigateway-auth/blob/main/maintenance/scheduled_maintenance.json (i.e. the default file) and I can manually edit all the existing files to have warning severity.

louise-davies (changes requested):
Works fine - just need to figure out what we want to do re: undefined `severity` on existing machines. Also, I fixed some custom CSS we added to toasts to make them more accessbile to apply to info and success toasts

src/state/actions/scigateway.actions.tsx
```diff
@@ -439,14 +439,17 @@ export const configureSite = (): ThunkResult<Promise<void>> => {
       provider
         .fetchScheduledMaintenanceState()
         .then((scheduledMaintenanceState) => {
+          if (scheduledMaintenanceState['severity'] === undefined) {
+            scheduledMaintenanceState['severity'] = 'warning';
+          }
```
kaperoo, line 444: The `severity` prop in scheduledMaintenanceState now defaults to `warning`.

louise-davies (approved):
Works fine, and I've tested removing the severity state on the server and the toast properly shows a warning toast

Issue #1312, opened 2023-09-29
Force refresh when site is taken out of maintenance mode


louise-davies:
**Description**:
This helps when doing a bugfix upgrade - we can ensure that no users have the old version of the code in their browser loaded after a certain period of time (i.e. how long SG checks for maintenance). For other circumstances - forcing a reload after maintenance mode is fairly non-intrusive - the only thing that is lost is any caching that the plugins may do. For this to work, we'd need to bump up the frequency of how often we check for maintenance mode changes - probably every 5 minutes is more suitable than every hour.

**Acceptance criteria**:
- [ ] When SG detects it's gone from maintenence mode true to false, it forces a page refresh. Easiest place to do this is probably the middleware I think. We don't need to use react-router reload method if it's more complex, basic `window.location.reload()` will do.
- [ ] Maintenance mode is checked for more frequently

PR #1315, opened 2023-10-09
Check for maintenance state every 5 minutes and force refresh if changed

kaperoo:
## Description
Check for maintenance status every 5 minutes. If the status has changed from true to false - force refresh the page.

## Testing instructions
Updated unit tests in `App.test.tsx` to check for page reloads.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to #1312

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -45,7 +45,7 @@ describe('App', () => {
     expect(screen.getByText('Loading...')).toBeInTheDocument();
   });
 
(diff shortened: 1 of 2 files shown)


louise-davies (approved):
Functionality works well for force-refreshing a users instance of SG. This doesn't reload the admin who unsets the maintenance mode page - but that's fine and they should know to close any instances of SG they had open before any update.

PR #1376, opened 2024-03-01
Increase toast functionailty #355

joshuadkitenge:
## Description
Allowed the toast to be shown if the show variable in the payload is "true".
Allowed all the toasts to be remove using esc

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes /ral-facilities/inventory-management-system/issues/355


src/state/middleware/scigateway.middleware.tsx
```diff
@@ -169,9 +169,11 @@ export const listenToPlugins = (
 
             // If "instant" is not used, by default we only show "warning" and "error" messages.
             if (pluginMessage.detail.payload.severity !== undefined) {
-              const { severity, message } = pluginMessage.detail.payload;
+              const { severity, message, show } = pluginMessage.detail.payload;
```
joelvdavies, line 172: Could you add a comment just to explain what this is for. Also should the name be more like `showSuccessToasts`/`showAllToasts` as giving a value of false will have no effect on errors. I guess it could even be changed to show only errors when show is `undefined` and otherwise just use that value so its configurable for all, in that case a name change wouldn't be necessary as it would just be an override.

src/state/middleware/scigateway.middleware.tsx
```diff
@@ -169,9 +169,14 @@ export const listenToPlugins = (
 
             // If "instant" is not used, by default we only show "warning" and "error" messages.
             if (pluginMessage.detail.payload.severity !== undefined) {
-              const { severity, message } = pluginMessage.detail.payload;
+              const { severity, message, showAllToasts } =
+                pluginMessage.detail.payload;
 
-              if (severity !== 'success') {
+              if (showAllToasts !== 'true' && severity !== 'success') {
+                // Shows all toasts messages expect for the success the toasts messages
```
joelvdavies, line 176:
```suggestion
                // Shows all toasts messages except for the success messages
```

joshuadkitenge:
I couldn't get the dispatchEvent working on cypress so I trigger a plugin error and clear that with escape for the e2e test

src/pageContainer.component.tsx
```diff
@@ -19,6 +20,26 @@ const PageContainer = (): React.ReactElement => {
   const theme = useTheme();
   const isViewportMdOrLarger = useMediaQuery(theme.breakpoints.up('md'));
 
+  React.useEffect(() => {
+    const clearToasts = (): void => {
+      toastr.clean();
```
joelvdavies, line 25: Is it not possible to mock `toastr` with `jest.mock`, and then check this is called to cover these lines?

joshuadkitenge: No, all of the configuration of the toasts (listening to the plugins) is done in the app.tsx

PR #1380, opened 2024-03-12
Migrate from CRA to Vite #1379

### LINKED ISSUE #1379: Migrate CRA to Vite
Create react app has not been maintained since April 2022. Similar to https://github.com/ral-facilities/inventory-management-system/pull/348 but will not require the MSW v2 changes. Should also replace the eslint plugin similarly to https://github.com/ral-facilities/inventory-management-system/pull/372.

joelvdavies:
## Description

Replaces CRA with Vite. Used the default settings that create-vite would have generated in the tsconfig. The only real change caused by that is a requirement of underscores on some unused variable & I removed the props on the footer as they weren't either (`noUnusedParameters`).

## Notes
- `react-redux-toastr` was incompatible (it used jsx in .js file) using old version, so upgraded v7.6.2 -> v7.6.13 which fixed
- Had an issue with `clearInterval` in `routing.component.test.tsx` not being defined which was fixed by https://github.com/testing-library/react-testing-library/issues/1198
- Timers behave slightly differently to before, requiring `{ shouldAdvanceTime: true }` in some `useFakeTimers`
- Unlike https://github.com/ral-facilities/inventory-management-system/pull/348, images could remain inside the src folder as it is not built in library mode and are exported correctly to dist/assets as a result
- Currently an issue with App.test.tsx due to the axios `__mock__` not having `expect.getState()` assigned as expected - fixed by mocking inside the test file itself (the token storage mock also needed to be hoisted to run correctly as it needed to be done before imports - see description here https://vitest.dev/api/vi#vi-hoisted)
- Like on https://github.com/ral-facilities/inventory-management-system/pull/348 cypress-failed-log has been removed
- Replaces  `eslint-config-react-app` with `eslint-plugin-react`, `eslint-plugin-testing-library` and `eslint-plugin-react-hooks` like https://github.com/ral-facilities/inventory-management-system/pull/372
- Removes `react-app-polyfill` and `custom-event-polyfill` (IE is no longer supported anyway, also see https://vitejs.dev/guide/build#browser-compatibility, we could use @vitejs/plugin-legacy if required)
- Uses `browserslist-to-esbuild` to add support for browserslist (View the currently targeted builds using `npx browserslist-to-esbuild`)

> [!WARNING]  
> The `index.html` has moved out of the `public` folder into the root, so `docker-ims` needs to be updated on gitlab as the title can no longer be replaced as it used to be, and instead will be in `/dist` after being built. Likely also effects any ansible config.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test against other front ends

## Agile board tracking
Closes #1379


src/adminPage/adminPage.component.tsx
```diff
@@ -59,7 +59,10 @@ const AdminPage = (props: AdminPageProps): ReactElement => {
       location.pathname.startsWith(adminRoutes[key])
     ) ??
       (props.adminPageDefaultTab &&
-      adminRoutes.hasOwnProperty(props.adminPageDefaultTab)
+      Object.prototype.hasOwnProperty.call(
```
louise-davies, line 62: I don't like this call signature. Can we switch to using `Object.hasOwn` instead? Browser compatibility seems good since we're dropping IE with this PR anyway.

src/mainAppBar/pageLinks.component.tsx
```diff
@@ -36,7 +35,8 @@ function PageLinks(): JSX.Element {
 
   function navigateToAdminPage(): void {
     const targetRoute =
-      adminPageDefaultTab && adminRoutes.hasOwnProperty(adminPageDefaultTab)
+      adminPageDefaultTab &&
+      Object.prototype.hasOwnProperty.call(adminRoutes, adminPageDefaultTab)
```
louise-davies, line 39: Change to `.hasOwn`

vite.config.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  // Allow hot reloading of json files in public folder when in development
+  if (env.NODE_ENV === 'development') plugins.push(jsonHMR());
+
+  return {
+    plugins: plugins,
+    server: {
+      port: 3000,
+      // Don't open by default as Dockerfile wont run as it can't find a display
+      open: false,
```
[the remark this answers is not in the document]
louise-davies: In fact, setting this to false doesn't actually mean anything, as currently it opens up for me just fine as it is!

louise-davies (commented):
Seems to be working ok, need to test with e.g. OG properly but I'll wait until my previous comments are addressed for a final review

src/adminPage/adminPage.component.tsx
joelvdavies: Yeah me neither, thanks for that 😄

vite.config.ts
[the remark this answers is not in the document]
joelvdavies: Yeah this is from IMS - in the package.json we actually use `vite dev --open` which overrides this which is why it still works for dev. This just overwrites the value. Thinking about it though I never needed to do this, as I could have just used `--open false` in the command in the docker container anyway so I will update them both and remove this from OG too.

louise-davies (dismissed):
LGTM and it's working well with OG. However, I don't want to merge this yet before we're ready to make the necessary deployment changes (e.g. scigateway-ansible).

I can probably make a branch for the changes for `scigateway-ansible` to support this work and merge it in once we are ready to switch to v3.x.x in production.

I've created an issue in scigateway-ansible: ral-facilities/scigateway-ansible#10

joshdimanteto (dismissed):
LGTM, tested with IMS

joelvdavies:
Note: Vite v6 support added to react router v7 (So still using v5 here)

joshdimanteto (dismissed):
LGTM, tested with IMS

package.json
louise-davies, line 1: Based on the commit name, it seems like you wanted to change the browserslist config to default but that has not been committed

joelvdavies: Thanks, vscode must have been slow at saving it.

louise-davies (changes requested):
Just realised there's a lingering reference to `build` instead of `dist` - in `.github/workflows/release-build.yml`, on this step:

```
 - name: Create tarball
        run: |
          mv build scigateway-$TAG_NAME
          tar -czf scigateway-$TAG_NAME.tar.gz scigateway-$TAG_NAME
```

We need to switch the `mv build` to `mv dist` as otherwise the built code won't get bundled for the release tar

louise-davies (approved):
LGTM - I have deployed this on `datagateway-dseg` and it works well :)

joshdimanteto (approved):
Tested with operations gateway and IMS. LGTM

louise-davies:
@joelvdavies I think this is ready to go in now that I've tested deploying it and the corresponding DataGateway PR got merged

PR #1382, opened 2024-03-22
Pin docker base images to SHA

VKTB:
## Description
This PR pins the base docker images in the Dockerfiles to commit SHAs so that Renovate (or Dependabot if Renovate doesn't work) can create PRs each time there is a change in the upstream base images that the base images are set to use as new SHAs are generated on every upstream change. Not sure if Renovate will also update the OS versions but if it doesn't then we will have to do the Alpine version updates manually for now.

Please note that I have pinned it to an older SHA to verify that Renovate (or Dependabot if Renovate doesn't work) will create a PR.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Login to Harbor
         uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
         with:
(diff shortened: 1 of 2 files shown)


joelvdavies (approved):

VKTB:
Renovate did create a PR to update the SHA but it did not bump the OS version so we will have to do this manually https://github.com/ral-facilities/scigateway/pull/1384

PR #1393, opened 2024-05-20
Fix error colour inconsistency for IMS

joelvdavies:
## Description
Resolves issue here: https://github.com/ral-facilities/inventory-management-system/issues/465.

This was caused because we override the error colours for various components but not generally when using the error colour directly. This adds in the override so they are all consistent. This could potentially have an undesired effect on datagateway if using `color="error"`, but I don't think it is being used anywhere I could find, and will only be going in the react-18 branch for the time being anyway,

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes ral-facilities/inventory-management-system#465

### DIFF
--- a/src/theming.tsx
+++ b/src/theming.tsx
@@ -72,7 +72,7 @@ interface ThemeColours {
 
   /* Lighter colours */
   lightBlue: string; //Used for lighter coloured buttons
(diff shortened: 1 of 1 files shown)


joshuadkitenge (approved):
LGTM

PR #1420, opened 2024-11-18
Add docker image push when tags pushed #1390

### LINKED ISSUE #1390: Update Docker job to create containers for tag & use org.opencontainers.image.source annotation
**Description**:
Currently, the docker job runs on pull requests & main & dev branches only, which means we don't have a proper pipeline to create Docker images when we tag a release. As part of addressing this, we should consider adding the org.opencontainers.image.source annotation (see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file?learn=dependency_version_updates&learnProduct=code-security#docker) - this requires the repo tags matches the docker images tags hence the first part of this issue is a prerequisite.

We should do this for all of our repos, just having the issue in SG to track it generally.

**Acceptance criteria**:
- [ ] Docker images are published to Harbor for tags (corresponding to releases)
- [ ] The org.opencontainers.image.source annotation is added to the Docker containers.

joelvdavies:
## Description
Adds a docker release action that builds and pushes the docker image when a tag is pushed. 

### Notes
- I tested the `docker/metadata-action` part of this new action on my own repo and found the following behaviour:

| GitHub Tag | Docker Image Tags |
| ------------- | ------------- |
| v1.0.1  | 1.0, 1.0.1, v1.0.1 |
| v1.0.1-dev | 1.0.1-dev, v1.0.1-dev |
| 1.0.1 | 1.0, 1.0.1 |
| 1.0.1-dev | 1.0.1-dev |
| docker-test | docker-test |

- We use the additional `type=semver,pattern={{major}}.{{minor}}` tag on IMS so our docker compose doesn't need to be updated for a patch update and can instead just point to `1.0` so I have done the same here

- `latest` is disabled here as it would otherwise always be added with tags that do not follow the semver format and we don't use it at the moment anyway. If it is needed in the future will likely need a separate workflow to filter the tag as regex.

- Disables docker builds from being pushed in PRs and branches instead only doing the build to ensure it can perform the docker build successfully prior to a release.

- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file?learn=dependency_version_updates&learnProduct=code-security#docker Suggests the tag must be the same for both the repo and docker image to function. As a result I think we may have to remove the `v` when doing releases though I am not 100% sure. It also reduces the duplication as in the table above.

- We still publish images under the datagateway project. This should probably be changed at some point in the future.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check labels and annotations make sense in the CI (see the `Extract metadata (tags, labels, annotations) for Docker` step)

## Agile board tracking
Closes #1390

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -38,32 +38,32 @@ jobs:
         run: yarn e2e
 
   docker:
-    # This job triggers only if all the other jobs succeed. It builds the Docker image and if successful,
-    # it pushes it to Harbor.
+    # This job triggers only if all the other jobs succeed. It builds the Docker image to ensure it builds correctly.
     needs: [test]
     name: Docker
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout repo
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Login to Harbor
-        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
         with:
           registry: ${{ secrets.HARBOR_URL }}
           username: ${{ secrets.HARBOR_USERNAME }}
           password: ${{ secrets.HARBOR_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


louise-davies (approved):
LGTM - if we keep tagging with the `v` prefix, I think it just means to get automatic updates via github dependabot the user will have to use the `v` prefix as well which I think is fine. We can always switch to no `v` if it gets annoying and recreate old tags.

PR #1442, opened 2025-06-04
Allow for authenticated homepage #1440 & hide nav drawer if it's empty

### LINKED ISSUE #1440: Allow for a homepage to be authenticated
In OG, it would be nice as it's a SPA to allow it to specify it's URL as the homepage, but currently SG assumes that the homepage is unauthenticated. It would be good to remove that special logic, and instead rely on plugins using the new `unauthorised: true` value in their routing config (see https://github.com/ral-facilities/scigateway/pull/1317)

This likely will require some changes to config in IMS and DataGateway which currently use unauthenticated homepages - but I think that should be fine? Unless we have special logic for the homepage which assumes if the `unauthorised` key is empty then assume it is unauthenticated rather than the other way round?

louise-davies:
## Description
These changes help to make having a true SPA (e.g. like OG) work a lot better. This allows a homepage to be authenticated, so an app like OG can specify that it's main URL is the homepage. As a follow up to this, this means that in a SPA like OG the nav drawer is empty, so added some handling to hide the nav drawer if it will be empty.

Also, I noticed no e2e tests for the unauthorised plugin feature, so I updated the demo_plugin to add an unauthorised route & added an e2e test. Corresponding demo_plugin PR here: https://github.com/ral-facilities/scigateway-demo-plugin/pull/201

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] BREAKING CHANGE: requires apps like IMS & DG to add `unauthorised: true` to the route config of their homepage. Otherwise these apps should be unaffected
- [ ] Test with OG by setting the homepageUrl to `/og` - the nav drawer should disappear and the homepage should still be authenticated

## Agile board tracking
Closes #1440


joshdimanteto (commented):
LGTM. It works as expected when run in local development. I tried running it using docker-operationsgateway, but it didn’t work. I built a new image for this branch and updated the docker-compose.yml to use that image and the relevant config (setting.json), but the drawer still appeared.

joshdimanteto (approved):

PR #1490, opened 2025-07-22
Add OIDC support #494

### LINKED ISSUE #494: Support OIDC authentication via ICAT authenticator
**Description**:
Brought up by Rolf, but we will eventually need to add an OIDC workflow to SciGateway to allow single sign on methods. This can perhaps also work with the ISIS user office.

We need more info on what is required, but thought it's useful to document the feature request from Rolf.

**Acceptance criteria**:
- [ ] Support for OIDC login to ICAT

louise-davies:
## Description

Adds support for OIDC via the icat authenticator, and adds 2 new authenticators: `oidc` and `password_and_oidc` - the former is for purely OIDC and the latter allows for both user/pass and OIDC( e.g. OG)

Also, majorily refactored LoginPage and the auth providers so that the fetching of e.g. ICAT mnemonics is now in the authenticators not LoginPage. Plus extracted out a lot of common code to `baseAPIAuthProvider` e.g. verify and refresh, OIDC login stuff, maintenance endpoints etc.

Only other big notable for testing is that I added a docker compose to set up a suitable keycloak e.g. for e2e tests, but since I added the code to `auth-server.js` this can also be used to test SG in dev mode with e.g. either the `oidc` or `userpass_and_oidc` authenticators

Also, there are breaking changes due to how OG-api and sg-auth handled the maintenance endpoints
- They now expect PUT requests rather than POST, which affects OG-api
- They now expect the token to be in the header and the request body to be the new state, affecting sg-auth

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Verify OIDC workflow works with new authenticators
- [ ] Verify OIDC workflow works with ICAT authenticator (will need to speak to me about setting up a suitable sg-auth server)

## Agile board tracking
Closes #494


cypress/e2e/login.cy.ts
```diff
@@ -439,6 +439,34 @@ describe('Login', () => {
...
+
+      // redirect?
+      // cy.origin()
+      cy.get('@originUrl').then((url) => {
+        cy.url().should('include', url);
+      });
+
+      cy.contains('Sign out').should('exist');
+      cy.get('[aria-label="Open user menu"]').should('be.visible');
+    });
```
joelvdavies, line 468: This is repeated in the describe below.

package.json
```diff
@@ -32,6 +32,7 @@
     "i18next-http-backend": "2.7.3",
     "js-cookie": "3.0.1",
     "jsonwebtoken": "9.0.0",
+    "jwks-rsa": "3.2.0",
```
joelvdavies: Should this be a dev dependency as its only used in the auth-server?

server/auth-server.js
```diff
@@ -83,6 +84,80 @@ app.post(`/login`, function (req, res) {
...
+  res.cookie('scigateway:refresh_token', refreshToken, {
+    httpOnly: true,
+    secure: process.env.HTTPS,
+    sameSite: 'lax',
+    maxAge: 604800,
+  });
+  res.status(200).json(accessToken);
+});
+
+// Fetch Maintenance State
```
joelvdavies: I don't think this comment makes sense?

src/authentication/baseAPIAuthProvider.tsx
```diff
@@ -0,0 +1,280 @@
...
+    })
+    .catch(() => {
+      log.error('Unable to fetch OIDC providers');
+      document.dispatchEvent(
+        new CustomEvent('scigateway', {
+          detail: {
+            type: NotificationType,
+            payload: {
+              message:
+                'It is not possible to authenticate you at the moment. Please, try again later',
```
joelvdavies:
```suggestion
                'It is not possible to authenticate you at the moment. Please, try again later.',
```

src/authentication/baseAPIAuthProvider.tsx
```diff
@@ -0,0 +1,280 @@
...
+              severity: 'error',
+            },
+          },
+        })
+      );
+      return [];
+    });
+}
+
+// GENERATING CODE VERIFIER
```
joelvdavies: Just wondering where this comes from and its context in OIDC?

src/state/state.types.tsx
joelvdavies, line 1: Unrelated to this PR but is the Plugin interface up to date? I didn't think unauthorised is in the plugin config but the routes?

joelvdavies (commented):
I think the description needs updating as it references `password_and_oidc` which is now `userpass_and_oidc`.

The existing authenticators worked well for me for local instances of IMS, DG and OG. I also tried username and password using `userpass_and_oidc` for OG which worked as well. The tests all worked locally for me too.  Have not tried OIDC directly yet.

joelvdavies (commented):
Just tried the oidc providers with the local keycloak instance and auth server which worked well 🥳 . I assume it would be tricky in practice to get a sensible username to display here?
image

cypress/e2e/login.cy.ts
louise-davies: They are meant to be different test cases, one is testing the plain OIDC auth and one is testing it working nicely with the `autoLogin` feature - however I forgot to rename the second test!

src/state/state.types.tsx
louise-davies: What do you mean by this? both the `RegisterRoutePayload` and the `PluginConfig` have `unauthorised` in them?

louise-davies:
> Just tried the oidc providers with the local keycloak instance and auth server which worked well 🥳 . I assume it would be tricky in practice to get a sensible username to display here? image

it's up to the auth server to put something sensible in the username field - I didn't put too much thought into the `auth-server.js` implementation tbh but in proper implementations we would expect something more like the user's email or fed id yeah.

I've changed it to attempt to read the email from the token if it's there :)

src/state/state.types.tsx
joelvdavies: I mean here, https://github.com/ral-facilities/scigateway/blob/b92311be63b4835458742e39cad07cd78ff7543d/src/state/state.types.tsx#L13-L19. I didn't think it was in the plugin config in the settings.json just the routes in the individual plugin configs.

louise-davies: Ah, I think this may have been a holdover from when we had the idea of allowing a whole plugin to be unauthorised like this, but yes this option doesn't do anything so I'll remove it

joelvdavies (dismissed):
LGTM 🥳

joshdimanteto (dismissed):
LGTM

louise-davies:
Added support for the updated interface from Alan, tested against sg-auth oidc branch and v0.7 of OG (this branch is deployed and working on OG preprod)

src/authentication/baseAPIAuthProvider.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+      if (!(await this.verifyOIDCNonce(id_token)))
+        throw Error('Nonce verification failed');
+
+      // TODO: sometimes login request fails here because it was too fast
+      // aka JWT iat is not yet "in the past"
+      // need to talk to backend people about it
+      await new Promise<void>((resolve, _reject) => {
+        setTimeout(() => resolve(), 1_000);
+      });
```
joelvdavies: I think this might have been addressed?
I can see Alan added a leeway on ral-facilities/scigateway-auth@aca9fab
Alex has on ral-facilities/operationsgateway-api@1ee23e6
and Viktor on https://github.com/ral-facilities/ldap-jwt-auth/pull/272/commits/073275370b2e59f71c24bae0a16f0de8a8a448fe.

joelvdavies (commented):
Just a query, worked well with IMS and https://github.com/ral-facilities/ldap-jwt-auth/pull/272.

src/authentication/baseAPIAuthProvider.tsx
louise-davies: Oh I must have forgot to push the commit 😂

joelvdavies (approved):

PR #1529, opened 2025-10-07
Release v3.1.0

louise-davies:
## Description
Quite a lot for a "minor" version aka Vite migration & OIDC but nothing backwards incompatible other than APIs not agreeing on what the format to edit maintenance modes should be - hence a minor version but could be convinced it needs to be a major.


joelvdavies:
I personally lean towards a major, it feels kind of excessive to me but https://github.com/ral-facilities/scigateway/pull/1442 was a breaking change for IMS.

louise-davies:
@joelvdavies ah I forgot that change was on here as well. I'll make it a major as it does completely overhaul the build system (even though it's non-breaking, it's a large change) and there are some breaking changes as well

PR #1542, opened 2026-01-12
Improve docker images & build workflow

louise-davies:
## Description
Improve the SG docker image so that you can more easily supply a settings.json via a volume and so that editing the page title can be done via an environment variable.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] I'm going to use these new images to deploy the new dev machine - I'll share the docker compose once that's done


.github/workflows/docker-release-build.yml
```diff
@@ -1,7 +1,18 @@
...
+      - develop
+    tags:
+      - '*'
+  workflow_dispatch:
+    inputs:
+      push-docker-image-to-harbor:
+        description: 'Push Docker Image to Harbor'
+        required: false
+        type: boolean
+        default: false
```
joelvdavies: Not used? (See DG)

Dockerfile
joelvdavies, line 1: Same warning on casing as DG PR. THere are also warnings on the use of env variables but I dont feel they are an issue.

docker/docker-entrypoint.sh
```diff
@@ -1,15 +1,40 @@
 #!/bin/sh -eu
 
+if [ ! -e /usr/local/apache2/htdocs/settings.json ]
+
+then
```
joelvdavies:
```suggestion
if [ ! -e /usr/local/apache2/htdocs/settings.json ]; then
```
With indenting in the if for readability?

joelvdavies (commented):
My comments are pretty much identical to DG PR. Conatianers for both built successfully, though I did not run them as will wait for the docker compose.

Dockerfile
louise-davies: I think it's flagging them because they contain the word `AUTH` - there's currently no way to whitelist a specific false positive and I feel disabling the check for the whole file isn't the right way to go so I think we just ignore it until we can ignore individual false positives

joelvdavies (dismissed):

joelvdavies (approved):

louise-davies:
I have deployed this on my new DGW dev cloud VM - will put the relevant docker compose onto gitlab in the docker-datagateway repo once I've sorted out SSL properly

PR #1545, opened 2026-01-27
Fix loading bug using more general solution

louise-davies:
## Description
I found another instance of the "loading bug" aka where single-spa thinks a plugin is mounted but the loading spinner is there. If you go to datagateway-dseg, go to one of the browse pages, and click either the search button or the cart button in the top left. This should 100% trigger the bug - I believe this is due to the navigation being in the plugin itself, so SG reacts slow enough to it that single-spa "mounts" the new plugin before the new plugin div is loaded.

This PR fixes the above bug, and I simplified the code to try and cover most use cases of the bug (aka rather than detecting specific circumstances, just look every time the route changes if we're on a plugin route, wait for the div to load and when it loads, if the preloader spinner is there retrigger single-spa loading)

## Testing instructions

- [ ] Review code
- [x] Check Actions build
- [x] Test it fixes the new instance of the loading bug - i.e. test that clicking the search or cart buttons in dataview doesn't trigger the bug
- [x] Test that switching from an admin route to a non-admin route doesn't trigger the bug (so go to the download admin page and then click on download in the sidebar) - I removed the specific fix for this but this new fix still covers this case

### DIFF
--- a/src/routing/routing.component.test.tsx
+++ b/src/routing/routing.component.test.tsx
@@ -371,106 +371,6 @@ describe('Routing component', () => {
     expect(history.location.pathname).toEqual('/logout');
   });
 
-  it('single-spa remounts a plugin when switching between admin and non-admin plugins via single-spa:before-no-app-change event', () => {
-    state.scigateway.authorisation.provider = new TestAuthProvider('logged in');
-    state.scigateway.siteLoading = false;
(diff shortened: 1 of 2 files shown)


joelvdavies (approved):
Fix working well for me in the cases described compared to develop 🥳

PR #1546, opened 2026-01-30
Improve post-login redirects

louise-davies:
## Description
There was an issue with the post-login redirects not working with `autoLogin` set to true so I fixed that here. However, that will only ever take you to the homepage as the `authorisedRoute` component will never redirect you to the login page with a `referrer` when auto-logged-in.

So I added a referrer when clicking the sign in button itself, which means we always attempt to put the user back on the page they were on before the sign in. I had to add some extra state `referredFrom` to not interfere with some of the `authorisedRoute` redirect logic.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Check that when `autoLogin` is true (so you need to point to an ICAT with an anon authenticator, dgw-dseg will do) you can login on a random page and after login you'll be redirected to the page you were on
- [x] Can also optionally test this behaviour with autologin off - e.g. go to the SG help page, sign in, it should take you back to the help page

### DIFF
--- a/cypress/e2e/login.cy.ts
+++ b/cypress/e2e/login.cy.ts
@@ -442,7 +442,10 @@ describe('Login', () => {
     });
 
     it('should be able to login via SSO after auto login and be displayed as logged in', () => {
-      cy.visit('/login');
+      // test it redirects us to plugin after login
+      cy.visit('/plugin1');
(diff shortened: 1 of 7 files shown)


joshdimanteto (approved):
LGTM tested with opsgateway and IMS as well

PR #1560, opened 2026-02-25
Ensure failing to fetch OIDC config details doesn't block logging in with other authenticators

louise-davies:
## Description
Change `initialiseOIDCProviders` to catch any errors `fetchOIDCConfig` throws and filter them out from the resulting array. This means a) we no longer block login if no OIDC authenticators work and b) if one OIDC provider fails then the others still work. Also improve the error messages to be more specific re: fetching OIDC config

This was noticed by Sabah yesterday as ORCiD had some maintenance yesterday that caused this issue

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] On datagateway-dseg I have 2 OIDC providers set up - I can configure one of them to always error to help test this change - let me know

### DIFF
--- a/src/authentication/baseAPIAuthProvider.test.tsx
+++ b/src/authentication/baseAPIAuthProvider.test.tsx
@@ -409,7 +409,9 @@ describe('Base API auth provider', () => {
           })
         );
 
-        await testBaseAPIAuthProvider.initialiseOIDCProviders();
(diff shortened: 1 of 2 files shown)


joelvdavies (approved):
Worked well for me, tested with a single and two providers with `ldap-jwt-auth`.

PR #1645, opened 2026-06-10
Add some translation related features

louise-davies:
## Description
These changes we want for the OG Gemini instance

- Allow userpass_and_oidc userpass authenticator to be translated
- Allow custom login error messages based on error code via context via translation
- Allow not providing a dont-have-an-account-sign-up-now translation which doesn't render that part. Also optimise this translation to be all in one without a link, but ensure backwards compatibility

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Test that removing the `dont-have-an-account-sign-up-now` removes the divider and translation
- [x] Test that adding a custom error code error message works (formatted like `login-error-msg_ERRORCODE` so e.g. `login-error-msg_403`

## Agile board tracking
Closes DSEGOG-544, DSEGOG-518


joelvdavies (commented):
Did you say the changes are supposed to be backwards compatible? When I use the old default.json I get

Image

Otherwise followed the testing instructions and worked for those cases.

louise-davies:
Ah yeah forgot that that is not backwards incompatible. I can probably add a built in default into the code.

joelvdavies (approved):

